PR #2, opened 2021-01-19
add top level Makefile

obnoxxx:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,3 @@
+build:
+	$(MAKE) -C images/samba build
+.PHONY: build


phlogistonjohn (approved):
Seems OK, maybe in the future it would be preferable to only have one makefile in the root rather than have to descend into subdirs, but I have no issue with this as a starting point.

obnoxxx:
> Seems OK, maybe in the future it would be preferable to only have one makefile in the root rather than have to descend into subdirs, 

Why is that? Just curious because I usually prefer to have Makefiles in the various directories.

phlogistonjohn:
Personally, if the project is using make more of a "front door" and not to actually build stuff like a C project I usually just like to have everything in one place. I have also heard a lot of people "diss" recursive makefiles but I would have to read stuff to remind myself why.

PR #6, opened 2021-01-20
Add a simple test

obnoxxx:
This adds a simple test for the container.
The test runs the container, and executes smbclient to simply list the shares.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,15 @@
+CONTAINER_CMD ?=
+ifeq ($(CONTAINER_CMD),)
+	CONTAINER_CMD:=$(shell docker version >/dev/null 2>&1 && echo docker)
+endif
+ifeq ($(CONTAINER_CMD),)
+	CONTAINER_CMD:=$(shell podman version >/dev/null 2>&1 && echo podman)
+endif
+
 build:
 	$(MAKE) -C images/samba build
 .PHONY: build
+
+test: build
+	CONTAINER_COMMAND=$(CONTAINER_CMD) hack/test-samba-container.sh
+.PHONY: test
--- a/hack/test-samba-container.sh
+++ b/hack/test-samba-container.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+LOCAL_TAG="${LOCAL_TAG:-centos8:samba}"
+
+if [ -z "${CONTAINER_CMD}" ]; then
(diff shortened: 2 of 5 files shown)


obnoxxx:
Failure in running the container due to `slirp4netns` not found. trying with `--network=none` for a start.

Makefile
```diff
@@ -1,3 +1,7 @@
 build:
 	$(MAKE) -C images/samba build
 .PHONY: build
+
+test: build
+	@hack/test-samba-container.sh
```
phlogistonjohn, line 6: Why hide what make is running?

hack/test-samba-container.sh
```diff
@@ -0,0 +1,40 @@
...
+
+LOCAL_TAG="${LOCAL_TAG:-centos8:samba}"
+
+TMPDIR="$(mktemp -d)"
+rc=$?
+
+if [ $rc -ne 0 ]; then
+	echo "Error creating temporary directory"
+else
+	CONTAINER_ID="$(podman run --network=none --name samba \
```
phlogistonjohn, line 11: do you want to change it to $CONTAINERCMD or something and allow specifying either podman or docker? I use podman by default but I think docker is still more common. Or perhaps save that for a follow up pr?

obnoxxx: Right, the Makefile still directly uses podman for building and tagging etc. I would want to expand that more generally to cover both. I think it can be done afterwards.

Makefile
obnoxxx: Agreed. Updated

hack/test-samba-container.sh
obnoxxx: OK, I did the docker build change first #7 ... :-)

obnoxxx:
Updated to work with docker and Podman. Based on top of #7.

images/client/Makefile
```diff
@@ -1,11 +1,19 @@
+CONTAINER_CMD ?=
```
phlogistonjohn, line 1: This is one reason a single makefile may be better. No need to repeat yourself like this. :-)

phlogistonjohn (approved):

images/client/Makefile
obnoxxx: I agree. exactly my thought when I did the patches... :grin: 
Room for more improvements...

PR #7, opened 2021-01-20
make it possible to build with docker instead of podman

obnoxxx:


phlogistonjohn (approved):

obnoxxx:
This is actually included in #6. :-)
I thought it might just realize that it's already in...

obnoxxx:
let's see what happens if i rebase locally and push over ;-)
(aka too bad that github only merges instead of rebasing when you hit the update branch button....)

PR #8, opened 2021-01-25
[RFC] samba: add a new dockerfile based on fedora and sambacc

phlogistonjohn:
The sambacc projects provides a script 'samba-container' meant to help
manage samba within a container environment. Fedora provides us
with some newer packages, etc.

Signed-off-by: John Mulligan <jmulligan@redhat.com>

PR #9, opened 2021-01-25
[RFC] makefiles: flatten and simplify

phlogistonjohn:
Make isn't really used to build anything and is more what I consider a
"familiar and friendly front door" for a project that helps newcomers
preform simple top-level actions in the repo easily. As such I don't
think recursive makefiles help a lot.

In addition, I have changed from some hard-coded values to ones in
variables that can be customized.

Finally, this change elimnates the extra "tag" step that was just
redundant with the --tag option to the 'build' command(s).


obnoxxx (changes requested):
In general this seems like a good improvement, specifically making hard-coded things configurable.

Personally, I am a fan of Makefiles for building etc. It avoids me having to remember all the commands needed for building... So even in subdirs, they can have a purpose. But I am ok with removing it and consolidate in the top lvl Makefile for now. 

A few points that we might want to change for the PR:

* It is based on PR #8, and I think that's not needed
* It silently swaps the build target from the centos based container to the fedora based one introduced in #8. I think this should be done as a conscious, explicit step instead.
* Getting rid of the separate tag step is fine (maybe a separate commit?), I mainly followed some steps for podman, and was confused what these really meant. Since it involves a quay.io component and the build command does not, I kind of assumed that this separate step is needed... :-)

obnoxxx:
note the "Container Image CI" check fails. But it does not actually fail at the build step but later at the run step:

```
CONTAINER_COMMAND=docker hack/test-samba-container.sh
63
Unable to find image 'centos8:samba' locally
```

So the `hack/test-samba-container.sh` script still has the `centos8:samba` tag hard coded and needs to be fixed. CI works! 😁 

But I don't actually understand how the build works since I don't know where the build.sh script comes from...

phlogistonjohn:
This PR has been completely re-worked to:
* Make changes in a more step-by-step manner, with smaller incremental commits
* Does not depend on PR #8  any more (the relationship will be reversed)
* Use clearer and cleaner variable naming (I hope)

Note that neither in this version or the previous could I test pushing to the canonical (default) quay.io repos as I don't have rights to them. Setting SERVER_REPO_NAME, for example, to a repo I have access to does work locally.

I would like to suggest changing the names of the canonical quay.io repos some time, but I think that's out of scope for this PR.

obnoxxx (approved):
Thanks for the rework. This is perfect!

/lgtm

PR #11, opened 2021-01-26
Use new org/repos under quay.io/samba.org by default

obnoxxx:
Moving away from the personal repo.
Created a new org "samba.org" and repos with more
systematic names under that org.
Use them by default.


phlogistonjohn (approved):
Sorry for the copy-n-paste mistake. LGTM.

PR #12, opened 2021-01-27
Push to quay after push to master

obnoxxx:
Based on #11.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -18,3 +18,20 @@ jobs:
       run: make build
     - name: Test the image
       run: make test
+
+  push:
+
+    # verify it passes build and test first
+    needs: build
+
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: log in to quay.io
+        run: podman login -u "${{ secrets.QUAY_USER }}" -p "${{ secrets.QUAY_PASS }}" quay.io
+      - name: push server image
+        run: make push-server
+      - name: push client image
+        run: make push-client


.github/workflows/container-image.yml
```diff
@@ -18,3 +18,20 @@ jobs:
...
+    if: github.event_name == 'push'
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: log in to quay.io
+        run: podman login -u "${{ secrets.QUAY_USER }}" -p "${{ secrets.QUAY_PASS }}" quay.io
+      - name: push server image
+        run: make push-server
+      - name: push client image
+        run: make push-client
```
phlogistonjohn, line 37: It's no biggie but would it make more sense to just push everything in one step like: `make push-server push-client`?

.github/workflows/container-image.yml
```diff
@@ -18,3 +18,20 @@ jobs:
...
     - name: Test the image
       run: make test
+
+  push:
+
+    # verify it passes build and test first
+    needs: build
+
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push'
```
phlogistonjohn, line 28: Does this include merges post-pull-request?

obnoxxx: yes, direct push and pushes/merges of a PR (is my understanding)

.github/workflows/container-image.yml
obnoxxx: dunno. there is no synergy between the two, right? so this way, we'll see more separate steps in the logs here. 
If you want to club them, I suggest to add a make target  "push" just like the "build" target... ;-)

.github/workflows/container-image.yml
obnoxxx: of course we could also create a separate workflow yaml just for the pushing, if that's more obvious. but my understanding is that this event_name directly relates to the `on:` section above

phlogistonjohn: It just wasn't too obvious to me w/o digging through the github actions documentation. We can see if it works in a practical manner soon. :-)

phlogistonjohn (approved):

obnoxxx:
gosh, why the conflicts....
i'll push a rebase

phlogistonjohn:
I approved it. But it needs a manual rebase?

obnoxxx:
for me, it does not even need rebasing: locally checking it out, it is cleanly based on top of the master branch...

obnoxxx:
heh, that's due to the merge commit added by the update button :man_facepalming:

obnoxxx:
rebased properly

PR #13, opened 2021-01-27
makefile: fix pushing with docker cli

phlogistonjohn:
Unlike podman, docker can not push from a local name to a remote name,
and only takes one argument: the remote name. Since I only have podman
locally I mistakenly changed the makefiles to require the podman
approach. This fixes that by applying both local and remote names at
build time and then pushing with only one argument.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,19 +25,19 @@ build: build-server build-client
 .PHONY: build
 
 build-server:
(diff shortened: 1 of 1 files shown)


obnoxxx (approved):
lgtm, even though I have to admit that I don't fully understand the docker thing.
the podman approach is just so much more logical ... :facepalm:

PR #14, opened 2021-01-27
github actions: enforce use of podman in the push action

obnoxxx:
The env has both docker and podman commands, and the
Makefile has some detection logic that currently prefers
docker.

We are using "podman login" in the yaml, so enforcing
the use of podman in the workflow to fix the post-merge
push action.


phlogistonjohn (approved):
This is probably fine on its own but it may end up running some work more than once (some in docker some in podman) and thus I recommend adding CONTAINER_CMD=podman to all the make invocations in the YAML (or via an env var?).
But this should work on its own for now.

PR #15, opened 2021-01-27
examples: add podman based AD joined winbind+smbd pod

phlogistonjohn:
Adds a script, config json template, and readme file to give a brief
into to running samba containers as a domain member.

### DIFF
--- a/examples/podman/README.md
+++ b/examples/podman/README.md
@@ -1,0 +1,40 @@
+
+# Podman Examples
+
+These examples demonstrate using samba container images with podman.
+
+## Domain Joined SMB Shares
+
+Included here is a shell script and a JSON config template that demonstrates
+using podman to run an smbd instance joined to Active Directory.  Do note that
+the Active Directory in question must already exist and permit new joins using
+the Administrator account. You *must* edit the provided files to match your
+domain/environment.
+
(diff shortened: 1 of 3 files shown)


examples/podman/README.md
```diff
@@ -0,0 +1,29 @@
+
+# Podman Examples
+
+These examples demonstrate using samba container images with podman.
+
+## Domain Joined SMB Shares
+
+Included here is a shell script and a JSON config template that
+demonstrates using podman to run an smbd instance joined to Active Directory.
```
obnoxxx, line 9: I think we might make it a little more clear that this examples requires an existing AD domain that the winbindd container will actually join with the credentials provided in the script.

obnoxxx (changes requested):
Great stuff!
Just one comment inline (can we make it more clear that we're joining an *existing* domain?)

examples/podman/README.md
phlogistonjohn: Sure. Will do.

phlogistonjohn: Done.

phlogistonjohn:
Added a section to the doc touching on the potential need to configure /etc/sub{uid,gid} files if running this with rootless podman.

obnoxxx (approved):
/lgtm

thanks!

PR #21, opened 2021-02-03
examples: add a yaml for smbd+winbind pod

phlogistonjohn:
This adds a basic, but functional, example of running a pod in kubernetes with both sbmd and winbind for use as a domain member smb server.

The YAML contains a ConfigMap, secret, and pod configuration. One must edit parts of the yaml (and the embedded JSON) to match the  user's domain.

This currently needs a new build of the samba server container to work thus it won't work without the commit bumping the version of sambacc in use. If you want I can split these into separate PRs. I wasn't sure if that was needed as this is "just" an example.

### DIFF
--- a/examples/kubernetes/sambadmpod.yml
+++ b/examples/kubernetes/sambadmpod.yml
@@ -1,0 +1,177 @@
+#
+# A Samba smbd + winbindd pod that automatically becomes a domain member.
+# This setup requires the config map "samba-conainer-config" to match
+# your test domain. It requires the Administrator password to be present
+# in the "ad-join-secret" secret.
+#
+# This configuration assumes you can reach your AD's DNS via the normal
+# Kubernetes DNS configuration. If not you may wish to look into the
+# customizing the Pod DNS settings or Kubernetes DNS customization.
+#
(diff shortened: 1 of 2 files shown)


examples/kubernetes/sambadmpod.yml
```diff
@@ -0,0 +1,177 @@
...
+# You can tweak the samba-sharedir volume under "volumes" at the end
+# of the pod yaml below. Use this to toggle between a temporary share
+# or a PVC backed share.
+#
+# NOTE: All settings below can be modified but those mentioned above
+# will get you a running container more quickly, feel free to start
+# there and then begin tinkering!
+#
+---
+# Configuration for thae samba domain member pod.
```
obnoxxx, line 26:
```suggestion
# Configuration for the samba domain member pod.
```

obnoxxx (changes requested):
small typo

examples/kubernetes/sambadmpod.yml
phlogistonjohn: Fixed.

obnoxxx (approved):
lgtm

PR #22, opened 2021-02-08
github actions: provide password via stdin

obnoxxx:
This is more secure than providing the password as a command line
argument to podman login.


obnoxxx:
because of the results and discussion in https://github.com/samba-in-kubernetes/samba-operator/pull/29, I think we can just close this. using the stdin is not helping here, afaict.

PR #24, opened 2021-02-10
images: use the sambacc container from quay.io/samba.org

obnoxxx:
Now that the sambacc projects is pushing its images to
the sambacc repo under quay.io/samba.org, let's use this
image repo instead of John's personal one. :-)


obnoxxx:
had used the wrong tag for the new image repo - fixed

phlogistonjohn (approved):

PR #26, opened 2021-02-16
kubernetes: update samba dm pod yaml to use init containers

phlogistonjohn:
The init containers are responsible for early env setup as well
as handling (if possible) and requiring domain joins.

This depends on PR #25 as the init containers require newer versions of sambacc.


obnoxxx (approved):
lgtm

PR #28, opened 2021-03-01
Add basic ad-server container

obnoxxx:
Files taken from https://hg.sr.ht/~phlogistonjohn/samba-ad-container

For a start, domain, users, etc, are all hard-coded.
We will generalize as we move on.

Co-authored-by: John Mulligan <jmulligan@redhat.com>

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -36,6 +36,8 @@ jobs:
       - name: push server image
         # note: forcing use of podman here, since we did podman login above
         run: make CONTAINER_CMD=podman push-server
+      - name: push ad-server image
+        run: make CONTAINER_CMD=podman push-ad-server
       - name: push client image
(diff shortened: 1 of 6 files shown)


phlogistonjohn (approved):
Looks like a good start to me.

I am not sure if we want add some documentation that notes that currently the ad container must be run as a privileged container. There are some small clean ups we could do too. Both of these could be done in follow up prs.

obnoxxx:
> Looks like a good start to me.
> 
> I am not sure if we want add some documentation that notes that currently the ad container must be run as a privileged container.  There are some small clean ups we could do too. Both of these could be done in follow up prs.

Yeah, I realize that I forgot the doc update in the PR, but the main README file could use some updating anyway. So good to bring the docs and other improvements / cleanups in follow-up PRs.

PR #30, opened 2021-03-04
ad-server: fix populate script

obnoxxx:
It was erroneously using "range" instead of "seq"
in enumeration...


phlogistonjohn (approved):
Silly of me not have fixed that earlier. LGTM.

PR #31, opened 2021-03-04
github actions: re-arrange checks for more granularity

obnoxxx:


obnoxxx:
Added Makefile improvements (making build targets non-phony).
In order to take full advantage of this, and avoid rebuilds, we'd need to pass artifacts between jobs.

Makefile
```diff
@@ -44,9 +46,10 @@ push-ad-server: build-ad-server
 	$(PUSH_CMD) $(AD_SERVER_REPO_NAME)
 .PHONY: push-ad-server
 
-build-client:
+build-client: $(BUILDFILE_CLIENT)
+$(BUILDFILE_CLIENT): Makefile $(CLIENT_SRC_FILE)
 	$(BUILD_CMD) --tag $(CLIENT_NAME) --tag $(CLIENT_REPO_NAME) -f $(CLIENT_SRC_FILE) $(CLIENT_DIR)
-.PHONY: build-client
+	date -R >> $(BUILDFILE_CLIENT)
```
phlogistonjohn, line 58: Why date? Why not put the hex container id from the build command?
(Also the output of date -R is not machine friendly IMO)

Makefile
```diff
@@ -13,6 +13,7 @@ SERVER_DIR:=images/server
 AD_SERVER_DIR:=images/ad-server
 CLIENT_DIR:=images/client
 SERVER_SRC_FILE:=$(SERVER_DIR)/Dockerfile.fedora
+SERVER_SOURCES:=$(SERVER_DIR)/setuser.sh $(SERVER_DIR)/smb.conf $(SERVER_DIR)/smb.import.conf
```
phlogistonjohn, line 16: I don't know if we still use these files. It might be better to remove them if we can? Maybe we left them behind before sambacc was more mature? But I think sambacc should be the way forward (of course I am bound to think that :-))

obnoxxx: fixed for this one. The current dockerfile only copies `smb.conf`.  We can remove the other files (+their dockerfile) later.

Makefile
obnoxxx: good point. updated to write container id and not date. (not sure why the machine friendly format for the content would matter. also i missed the point with the container id first and just wanted to put *something* in - now I only put the container id).

obnoxxx:
@phlogistonjohn , @gd - if we want to merge this, we need to do it by override, I think, and then switch the branch settings over to remove the "build" target from the required checks and add the new checks...

phlogistonjohn (approved):
Seems reasonable. Thanks!

obnoxxx:
Thanks @phlogistonjohn  - merged and made the changes to the required checks.

PR #32, opened 2021-03-04
samba-ad-server: add example kubernetes deployment

obnoxxx:

### DIFF
--- a/examples/kubernetes/samba-ad-server-deployment.yml
+++ b/examples/kubernetes/samba-ad-server-deployment.yml
@@ -1,0 +1,61 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: samba-ad-server
+  labels:
+    app: samba-ad
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: samba-ad
+  template:
+    metadata:
+      labels:
+        app: samba-ad
+    spec:
+      containers:
+      - name: samba
+        # Pointing to locally built image,
+        # change here to use a registry.
+        image: samba-ad-container:latest
+        # Need imagePullPolicy Never for working with local images.
+        # Otherwise we get "ErrImagePull".
+        imagePullPolicy: Never
+        securityContext:
+          capabilities:
+            add: ["SYS_ADMIN"]
+        ports:
+        # https://wiki.samba.org/index.php/Samba_AD_DC_Port_Usage
+        - containerPort: 53
+          name: dns
+        - containerPort: 135
+          name: epm
+          protocol: TCP
+        - containerPort: 137
+          name: netbios-ns
+          protocol: UDP
+        - containerPort: 138
+          name: netbios-dgram
+          protocol: UDP
+        - containerPort: 139
+          name: netbios-session
+          protocol: TCP
+        - containerPort: 389
+          name: ldap
+        - containerPort: 445
+          name: smb
+          protocol: TCP
+        - containerPort: 464
+          name: kerberos
+        - containerPort: 636
+          name: ldaps
+          protocol: TCP
+        - containerPort: 3268
+          name: gc
+          protocol: TCP
+        - containerPort: 3269
+          name: gc-ssl
+          protocol: TCP
+        # need 49152-65535 for dynamic RPC ports
+        # but currently not possible to specify ranges


examples/kubernetes/samba-ad-server-deployment-local.yml
```diff
@@ -0,0 +1,25 @@
...
+      app: samba-ad
+  template:
+    metadata:
+      labels:
+        app: samba-ad
+    spec:
+      containers:
+      - name: samba
+        image: samba-ad-container:latest
+        imagePullPolicy: Never
```
phlogistonjohn, line 20: I'm confused as to how this "samba-ad-server-deployment-local.yml" is supposed to differ from the other one. Unless I am missing something the main difference is the imagePullPolicy, but I don't understand what that's meant to accomplish.

examples/kubernetes/samba-ad-server-deployment.yml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: samba-ad-server
+  labels:
+    app: samba
```
phlogistonjohn, line 6: Should this be 'samba-ad' for consistency?

examples/kubernetes/samba-ad-server-deployment.yml
```diff
@@ -0,0 +1,24 @@
...
+        app: samba-ad
+    spec:
+      containers:
+      - name: samba
+        image: quay.io/samba.org/samba-ad-server:latest
+        securityContext:
+          capabilities:
+            add: ["SYS_ADMIN"]
+        ports:
+        - containerPort: 445
```
phlogistonjohn, line 24: As this is an ad server I think that at the least we should also be exposing ports for dns, ldap and perhaps ldaps.

phlogistonjohn (changes requested):
The gist is good but I have a few suggestions and I want to better understand why there are two yamls that seem very similar being added.

examples/kubernetes/samba-ad-server-deployment.yml
obnoxxx: true, this was just a start. probably I should mark this Draft... ;-)

examples/kubernetes/samba-ad-server-deployment-local.yml
obnoxxx: The main point was using the locally built image instead of pulling the one from quay.io.
After some googling it seemed for it to work, I had to change the pull policy to Never. 🤷

obnoxxx: If we can get away with just one deployment - fine!

phlogistonjohn: OK, my thoughts are that since this is an example file, we can simply do one file, perhaps with comments added to explain some of the common things a person might want to change. 

FWIW, I usually just change the image to reference a container registry I run on my machine or my own repo on a public container registry. The former has the added advantage that its really much faster, but has the downside you must to extra configuration steps before it will work.

examples/kubernetes/samba-ad-server-deployment.yml
obnoxxx: updated

examples/kubernetes/samba-ad-server-deployment.yml
obnoxxx: updated with a TODO comment for now

examples/kubernetes/samba-ad-server-deployment-local.yml
obnoxxx: updated to only have one yaml. the one with the local image.

obnoxxx:
@phlogistonjohn - this is now being used in #34 where it is tested in the CI in k3d (k3s in docker). 😎

obnoxxx:
@phlogistonjohn - added the ports, but it's quite massive and there does not seem to be a way to specify a range.
I don't know whether we really need to specify all those ports.

examples/kubernetes/samba-ad-server-deployment.yml
phlogistonjohn: Heh. I think this is fine for now. Some of these might be considered "obsolete" but I don't know what applications and platforms use what. So I think this is good enough.

phlogistonjohn (commented):
I would prefer these patches get squashed together, but would be OK with it as is.

obnoxxx:
> I would prefer these patches get squashed together, but would be OK with it as is.

Yes, I meant to eventually squash, but wanted to add on top for initial comments. Will squash now.

examples/kubernetes/samba-ad-server-deployment.yml
obnoxxx: Yeah, we can still clean up later.

phlogistonjohn (approved):

PR #36, opened 2021-03-18
kubernetes examples: tweak ad deployment to use public image by default

phlogistonjohn:
I think that defaulting the AD server example deployment to the public
AD server image is probably the better default. There will be fewer
additional steps for new users to get an AD instance running in k8s.

FWIW, I configure cri-o on my scratch clusters to point at a registry
mirror I control, so I can pre-cache my images locally. This lets me
use many of the examples as-is.

Signed-off-by: John Mulligan <jmulligan@redhat.com>


obnoxxx (commented):
Initially, I had two almost identical yamls with precisely this diff between them. I went with the committed one to make sure that the CI against a PR will really test the changes to the container that are possibly made in the PR. If I am not wrong, with this proposed change, the CI would pull the image from quay, so *not* testing the PR-changes. Is there a way to fix this?

obnoxxx:
reworked the testing in #37 to use its own deployment yaml. After that is merged, we can get this in.

obnoxxx:
Now that #37 is merged, this can go in.

obnoxxx (approved):
lgtm

PR #37, opened 2021-03-18
Make ad server test use dedicated deployment file

obnoxxx:
This way, we can have the examples file to an available published image.


phlogistonjohn (approved):
LGTM. This is pretty cool! :-)

PR #38, opened 2021-03-28
Enhance ad server kubernetes level tests with member server setup

obnoxxx:


obnoxxx:
@phlogistonjohn - FYI, WIP PR with kubernetes level ad-server and ad-member test.

tests/test-remove-ad-server.sh
```diff
@@ -0,0 +1,16 @@
...
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+source "${SCRIPT_DIR}/common.sh"
+
+if [ ${KEEP} -eq 1 ]; then
+	echo "keeping ad server deployment (KEEP=1)"
+	exit 0
+fi
+
+echo "removing ad server deployment again..."
```
phlogistonjohn, line 12: Very nitpicky, but the word "again" here strikes me as unnecessary.

tests/test-remove-ad-server.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+
+source "${SCRIPT_DIR}/common.sh"
+
+if [ ${KEEP} -eq 1 ]; then
```
phlogistonjohn, line 7: Should this be named something more specific like KEEP_AD or KEEP_AD_DEPLOYMENT?

tests/test-samba-ad-server-kubernetes.sh
```diff
@@ -10,6 +10,14 @@ kubectl exec "${podname}" -- samba-tool domain info 127.0.0.1
 [ $? -eq 0 ] || _error "Error listing domain info"
 echo
 
+source ${SCRIPT_DIR}/test-deploy-ad-member.sh
+
+kubectl exec "${podname}" -c "smb" -- smbclient -N -L 127.0.0.1
+kubectl exec "${podname}" -c "winbind" -- wbinfo -t
+kubectl exec "${podname}" -c "smb" -- wbinfo -t
```
phlogistonjohn, line 17: I was momentarily confused by two "wbinfo -t" but then I realized it makes sense as the container running smbd needs to communicate with winbind and `wbinfo -t` can demonstrate that. So, perhaps a comment will save future readers like me from that confusion.

phlogistonjohn (commented):
General +1 to the gist of things here. But as it's still in draft I will look in more detail later.

phlogistonjohn:
While reading this I also realized we should do something about continuing to use "my" domain and filed issue #39

tests/test-remove-ad-server.sh
obnoxxx: right, I thought I had changed that...

tests/test-remove-ad-server.sh
obnoxxx: Well, I was thinking about this.
The current patch uses `$KEEP` in both `test-remove-ad-server.sh` and `test-remove-ad-member.sh`.
I could introduce two separate variables...

🤔

tests/test-samba-ad-server-kubernetes.sh
obnoxxx: exactly, that's the point. :-)
Will add a comment.

obnoxxx:
> While reading this I also realized we should do something about continuing to use "my" domain and filed issue #39

Right. Let me integrate into this PR a patch to first change the domain to something else.

tests/test-samba-ad-server-kubernetes.sh
obnoxxx: done

tests/test-remove-ad-server.sh
obnoxxx: done

tests/test-remove-ad-server.sh
obnoxxx: kept is as one var for now

obnoxxx:
updated and polished the PR.
Based on #40 now

obnoxxx:
@phlogistonjohn  -  updated the test to test the locally built image instead of the one from quay

phlogistonjohn (approved):
Looks good. Since nothing here is set in stone I don't think its worth nitpicking anything else. :-)

(Well, I did notice that one commit messages says "kubernets"... but I will approve anyway and leave it to you to either fix or not worry about a minor typo)

obnoxxx:
Ah, you don't know this kubernets? ... 😁 
I think I don't care.

Issue #39, opened 2021-03-29
AD: default to a safer example domain


phlogistonjohn:
Because I originally cooked up the scripts we use for the AD image as a quick and dirty prototype for my own use the scripts set up the samba AD server with a phony subdomain of a domain I currently have registered for my own personal use.
While it's good that I know that there will not be any conflicts we should probably start looking towards cleaning this up sooner, rather than later, and preventing something so linked to me from proliferating in the scripts and tests.

I did a quick search and found https://tools.ietf.org/html/rfc6761
Based on this I think we're best off defaulting to "<something>.test." as our domain for local, ephemeral, test domains. At the same time I can make either the whole string or the "<something>" bit based on an environment variable to start making the scripts more usable for other cases and reducing the amount of stuff that is hard coded.

Opinions, @obnoxxx ?

obnoxxx:
Yes, this is right.
How about `testdom.sink.test` ? ==> PR #40

PR #40, opened 2021-03-29
ad-server: change default domain to testdom.sink.test

obnoxxx:
Fixes: #39

### DIFF
--- a/images/ad-server/provision.sh
+++ b/images/ad-server/provision.sh
@@ -3,7 +3,7 @@
 set -e
 
 REALM="DOMAIN1.SINK.TEST"
-DOM="DOMAIN1"
+DOMAIN="DOMAIN1"
 PW="Passw0rd"
 
 #if ! grep -q $HOSTNAME /etc/hosts ; then
@@ -19,6 +19,6 @@ samba-tool domain provision \
     --dns-backend="SAMBA_INTERNAL" \
     --server-role=dc \
     --realm="${REALM}" \
-    --domain="${DOM}" \
+    --domain="${DOMAIN}" \
     --adminpass="${PW}"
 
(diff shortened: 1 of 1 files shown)


phlogistonjohn (approved):
Looks good, thanks!

images/ad-server/provision.sh
```diff
@@ -2,8 +2,8 @@
 
 set -e
 
-REALM="ZZZ-BEST.X.ASYNCHRONO.US"
-DOM="ZZZ-BEST"
+REALM="samba1.sink.test"
```
phlogistonjohn, line 5: Unfortunately, I think using lower case here causes an error with the provisioning tool. You'd naturally think the domain is case sensitive but this [samba wiki page](https://wiki.samba.org/index.php/Setting_up_Samba_as_an_Active_Directory_Domain_Controller) says it must be upper cased.

phlogistonjohn (changes requested):

images/ad-server/provision.sh
obnoxxx: This was not the error. It worked before with `testdom`. The problem was that `samba1` is also the netbios name of the domain controller. Updated...

phlogistonjohn:
Sorry if I seem overly picky about this, but if you ended up changing the dc name in the end, what's wrong with samba1 now for the domain name?

phlogistonjohn:
> > Sorry if I seem overly picky about this, but if you ended up changing the dc name in the end, what's wrong with samba1 now for the domain name?
> 
> As I may have explained (maybe elsewhere), I am usually a fan of giving test objects names that indicate what kind of object this is. I.e something with `domain` for a domain, and something with `dc` for a dc ...
> 
> You are right, that it is not needed here. If you prefer, I can remove that part and put it into a separate PR for discussion.

It's fine to leave it. I was just a bit confused about why both values ended up getting changed.

phlogistonjohn (approved):

PR #43, opened 2021-06-25
server image: make the the server image build more flexible

phlogistonjohn:
To make it simpler to develop and test changes to sambacc in the server image, allow the use of build args specifying the version/tag/hash and repo from which to build the sambacc wheel.

Split the package install RUN from installing sambacc RUN; by splitting the build this way the layer containing the system packages and that containing sambacc install can be independently cached. The (probably) less frequently changing packages layer may be reused more often than the more frequently changing (for now) sambacc layer.


raghavendra-talur (approved):

obnoxxx (approved):
/lgtm

PR #44, opened 2021-08-12
experimental ctdb support

phlogistonjohn:
These changes add the ctdb package to the container, which is pretty harmless on its own, as well as the most recent sambacc which can "manage" aspects of ctdb if you opt in to the experimental ctdb support.

I've included an example yaml that makes use of a StatefulSet to demonstrate how the most basic integration of ctdb into a samba pod under k8s might work.

I'm putting this out here in a half-baked state so that others have an opportunity to try it out and provide feedback.

### DIFF
--- a/images/server/Dockerfile.fedora
+++ b/images/server/Dockerfile.fedora
@@ -1,6 +1,6 @@
 FROM quay.io/samba.org/sambacc:latest AS builder
-ARG SAMBACC_VER=a2cfac7ae49e
-ARG SAMBACC_REPO=
+ARG SAMBACC_VER=b279787f9550
+ARG SAMBACC_REPO=https://github.com/samba-in-kubernetes/sambacc
 
 # the changeset hash on the next line ensures we get a specifc
 # version of sambacc. When sambacc actually gets tagged, it should
@@ -13,7 +13,7 @@ MAINTAINER John Mulligan <jmulligan@redhat.com>
 ENV SAMBACC_VERSION="0.1"
 
 COPY smb.conf /etc/samba/smb.conf
-RUN dnf install -y \
+RUN dnf install --setopt=install_weak_deps=False -y \
     findutils \
     python-pip \
     python3-jsonschema \
@@ -23,7 +23,10 @@ RUN dnf install -y \
     samba-winbind \
     samba-winbind-clients \
     tdb-tools \
+    ctdb \
     && dnf clean all \
+    && cp --preserve=all /etc/ctdb/functions /usr/share/ctdb/functions \
+    && cp --preserve=all /etc/ctdb/notify.sh /usr/share/ctdb/notify.sh \
     && true
 
 COPY --from=builder \
--- a/examples/kubernetes/samba-ctdb-dm-sset.yml
+++ b/examples/kubernetes/samba-ctdb-dm-sset.yml
@@ -1,0 +1,428 @@
+#
+# An example of running samba with ctdb replication as a kubernetes
+# StatefulSet.
+#
+# This is *highly experimental* and not meant for real use. Use only if you're
+# interested in helping test or develop samba-container/sambacc CTDB support.
+#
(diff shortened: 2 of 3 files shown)


phlogistonjohn:
I'll probably merge this next week if I get no responses/NACKs

images/server/Dockerfile.fedora
```diff
@@ -23,7 +23,10 @@ RUN dnf install -y \
     samba-winbind \
     samba-winbind-clients \
     tdb-tools \
+    ctdb \
     && dnf clean all \
+    && cp --preserve=all /etc/ctdb/functions /usr/share/ctdb/functions \
+    && cp --preserve=all /etc/ctdb/notify.sh /usr/share/ctdb/notify.sh \
```
obnoxxx, line 29: what's the purpose of these copies? Is the rpm wrong?

images/server/Dockerfile.fedora
```diff
@@ -13,7 +13,7 @@ MAINTAINER John Mulligan <jmulligan@redhat.com>
 ENV SAMBACC_VERSION="0.1"
 
 COPY smb.conf /etc/samba/smb.conf
-RUN dnf install -y \
+RUN dnf install --setopt=install_weak_deps=False -y \
```
obnoxxx, line 16: what's the reason for this added flag?

images/server/Dockerfile.fedora
```diff
@@ -1,6 +1,6 @@
 FROM quay.io/samba.org/sambacc:latest AS builder
-ARG SAMBACC_VER=a2cfac7ae49e
-ARG SAMBACC_REPO=
+ARG SAMBACC_VER=b279787f9550
+ARG SAMBACC_REPO=https://github.com/samba-in-kubernetes/sambacc
```
obnoxxx, line 3: What's the relevance of this added line in this commit?

obnoxxx (commented):
Looks good in general. A few questions inline. 
And the commit message mentions `Include ctdb, some of its supporting scripts and config...` - what scripts and config are you referring to? Stuff that's installed with the rpm?

images/server/Dockerfile.fedora
phlogistonjohn: I wouldn't say wrong per-se, but since we (can) mount an emptydir over /etc/ctdb we'd lose access to those files. This copy makes sure we preserve them in a place that we're unlikely to mount over in a container environment. That said I do think it would be better if ctdb rpm put those not-strictly-configuration files in /usr. :-)

images/server/Dockerfile.fedora
phlogistonjohn: We started pulling in more than we needed, while it wasn't much more I remembered that `--setopt=install_weak_deps=False ` is a good practice when  running dnf in containers.

images/server/Dockerfile.fedora
phlogistonjohn: Technically leaving `SAMBACC_REPO` blank is the same as setting it, but I started getting mixed up when working on stuff so I thought it would be better to spell it out explicitly.

phlogistonjohn:
> `Include ctdb, some of its supporting scripts and config...` - what scripts and config are you referring to? Stuff that's installed with the rpm?

Yeah, I think I was referring to the cp commands where I copy some of the scripts contained in /etc.

images/server/Dockerfile.fedora
obnoxxx: ah, cool, thanks. wasn't aware...

images/server/Dockerfile.fedora
obnoxxx: makes sense.

images/server/Dockerfile.fedora
obnoxxx: 👍

obnoxxx (approved):
lgtm

PR #50, opened 2021-10-07
Consume nightly built Samba RPMs

anoopcs9:


anoopcs9:
**test-ad-server-kubernetes** failed because we don't have AD DC components built with our nightly Samba RPMs. I'll enable and check.

Makefile
```diff
@@ -6,19 +6,29 @@ ifeq ($(CONTAINER_CMD),)
 	CONTAINER_CMD:=$(shell docker version >/dev/null 2>&1 && echo docker)
 endif
 
+BUILD_TRIGGER ?=
```
phlogistonjohn, line 9: Instead of having conditionals in the makefile, I suggest we add new targets. See below. I also find this variable name confusing as "trigger" makes it sound like it should be activating something extra.

Makefile
```diff
@@ -6,19 +6,29 @@ ifeq ($(CONTAINER_CMD),)
...
 BUILD_CMD:=$(CONTAINER_CMD) build $(BUILD_OPTS)
 PUSH_CMD:=$(CONTAINER_CMD) push $(PUSH_OPTS)
 
 SERVER_DIR:=images/server
 AD_SERVER_DIR:=images/ad-server
 CLIENT_DIR:=images/client
 SERVER_SRC_FILE:=$(SERVER_DIR)/Dockerfile.fedora
+ifeq ($(BUILD_TRIGGER),nightly)
+	SERVER_SRC_FILE:=$(SERVER_DIR)/Dockerfile.nightly
+endif
```
phlogistonjohn, line 20: I think it would be better to add new targets to the makefile. Something like `build-server-nightly-pkgs`.  It might end up slightly longer than what we have here, but I think it will be more discoverable (tab completion) and obvious than a variable. It would also prevent confusion with the contents of the "buildfile" when building both types of container on one system.

phlogistonjohn (changes requested):
The dockerfile looks good, however I think a slightly different approach with the makefile would be better. Let me know what you think.

Makefile
anoopcs9: I have made some changes.

images/server/Dockerfile.nightly
```diff
@@ -0,0 +1,52 @@
+FROM quay.io/samba.org/sambacc:latest AS builder
+ARG SAMBACC_VER=f6480c5861a5
```
phlogistonjohn, line 2: I think we should sync this with what's currently in Dockerfile.fedora.

phlogistonjohn (changes requested):
Sorry for taking so long on this. Feel free to ping me in the future if I take more than a week to reply to a PR.
Anyway, one small change requested but everything else looks OK. In the future I think there's stuff we can do to perhaps minimize some duplication but I'd rather get this in now and deal with that later.

PR #52, opened 2021-10-13
server: Use the correct hash for sambacc repo

spuiuk:
When building the server image, the git hash for the sambacc repo is
incorrectly set to an old commit. As a result, the server image on
quay.io/samba.org/samba-server:latest
contains an old outdated version of sambacc.

Use the master branch instead when building the server images.

Signed-off-by: Sachin Prabhu <sprabhu@redhat.com>

### DIFF
--- a/images/server/Dockerfile.fedora
+++ b/images/server/Dockerfile.fedora
@@ -1,5 +1,5 @@
 FROM quay.io/samba.org/sambacc:latest AS builder
-ARG SAMBACC_VER=b279787f9550
+ARG SAMBACC_VER=f6480c5861a56ef9d1ebb965aebb14732ec2690c
 ARG SAMBACC_REPO=https://github.com/samba-in-kubernetes/sambacc
 
 # the changeset hash on the next line ensures we get a specifc


spuiuk:
Looks like the build.sh file, a copy which exists at
https://github.com/samba-in-kubernetes/sambacc/blob/master/tests/container/build.sh
attempts to run

git checkout -b master master

Running the buils.sh script  with bash -x
./t.sh master https://github.com/samba-in-kubernetes/sambacc
+ set -e
+ python=python3
+ url=https://github.com/samba-in-kubernetes/sambacc
+ bdir=/var/tmp/build/sambacc
+ '[' https://github.com/samba-in-kubernetes/sambacc ']'
+ url=https://github.com/samba-in-kubernetes/sambacc
+ mkdir -p /var/tmp/build/
+ checked_out /var/tmp/build/sambacc
+ local d=/var/tmp/build/sambacc
+ [[ -d /var/tmp/build/sambacc ]]
+ [[ -d /var/tmp/build/sambacc/.git ]]
+ echo 'repo already checked out'
repo already checked out
+ cd /var/tmp/build/sambacc
+ '[' master ']'
+ update /var/tmp/build/sambacc master
+ local d=/var/tmp/build/sambacc
+ local node=master
+ [[ -d /var/tmp/build/sambacc/.hg ]]
+ git checkout -b master master
fatal: A branch named 'master' already exists.

Unless we decide to fix this script, we may have to rely on commit hashes instead of using a branch name.

spuiuk:
John, What is the history of this build.sh file? Are we ok modifying it so that we can use branch names instead of commit hashes or do you think we rather not change this script?

spuiuk:
or we use a tag name. instead of branch master.

PR #55, opened 2021-11-17
bump sambacc "version"

phlogistonjohn:
This sambacc version has improvements around logging and better dns-reg behavior wrt
ctdb enabled instances.

### DIFF
--- a/images/server/Dockerfile.fedora
+++ b/images/server/Dockerfile.fedora
@@ -7,7 +7,7 @@ ARG SAMBACC_REPO=https://github.com/samba-in-kubernetes/sambacc
 # be changed to use the tag.
 RUN /usr/local/bin/build.sh ${SAMBACC_VER} ${SAMBACC_REPO}
 
-FROM fedora
+FROM registry.fedoraproject.org/fedora:34
 
 MAINTAINER John Mulligan <jmulligan@redhat.com>
 ENV SAMBACC_VERSION="0.1"
--- a/images/server/Dockerfile.fedora
+++ b/images/server/Dockerfile.fedora
@@ -1,5 +1,5 @@
 FROM quay.io/samba.org/sambacc:latest AS builder
-ARG SAMBACC_VER=f6480c5861a56ef9d1ebb965aebb14732ec2690c
+ARG SAMBACC_VER=fd747527ed86d01821c4a3b72f1c4f893bb1ca36
 ARG SAMBACC_REPO=https://github.com/samba-in-kubernetes/sambacc
 
 # the changeset hash on the next line ensures we get a specifc
--- a/images/ad-server/Containerfile
+++ b/images/ad-server/Containerfile
@@ -1,4 +1,4 @@
-FROM fedora
+FROM registry.fedoraproject.org/fedora:34
 
 RUN dnf install -y samba-dc procps-ng /usr/bin/smbclient
 RUN rm -f /etc/samba/smb.conf
--- a/images/server/Dockerfile.nightly
+++ b/images/server/Dockerfile.nightly
@@ -1,5 +1,5 @@
 FROM quay.io/samba.org/sambacc:latest AS builder
-ARG SAMBACC_VER=f6480c5861a56ef9d1ebb965aebb14732ec2690c
+ARG SAMBACC_VER=fd747527ed86d01821c4a3b72f1c4f893bb1ca36
 ARG SAMBACC_REPO=https://github.com/samba-in-kubernetes/sambacc
 
 # the changeset hash on the next line ensures we get a specifc


images/server/Dockerfile.fedora
```diff
@@ -1,13 +1,13 @@
 FROM quay.io/samba.org/sambacc:latest AS builder
-ARG SAMBACC_VER=f6480c5861a56ef9d1ebb965aebb14732ec2690c
+ARG SAMBACC_VER=fd747527ed86d01821c4a3b72f1c4f893bb1ca36
```
anoopcs9, line 2: With https://github.com/samba-in-kubernetes/sambacc/pull/26 merged, are we still gonna stay with commit hash and not `master` branch?

images/server/Dockerfile.fedora
```diff
@@ -1,13 +1,13 @@
...
 ARG SAMBACC_REPO=https://github.com/samba-in-kubernetes/sambacc
 
 # the changeset hash on the next line ensures we get a specifc
 # version of sambacc. When sambacc actually gets tagged, it should
 # be changed to use the tag.
 RUN /usr/local/bin/build.sh ${SAMBACC_VER} ${SAMBACC_REPO}
 
-FROM fedora
+FROM registry.fedoraproject.org/fedora:34
```
anoopcs9, line 10: [samba-4.15.2-3.fc35](https://bodhi.fedoraproject.org/updates/FEDORA-2021-12af2614da), which have fixes for `smbclient` anonymous login issues, should be available in stable repositories soon(mostly eod). So let's get to 35 while we are at it.

tests/test-deploy-ad-server.sh
```diff
@@ -49,7 +49,12 @@ while [ $tries -lt 120 ] && [ $rc -ne 0 ]; do
 	echo -n "."
 done
 echo
-[ $rc -eq 0 ] || _error "Error: samba ad did not become reachable"
+if [ $rc -ne 0 ] ; then
+	echo -n "---- DEBUG -----"
+	kubectl exec "${podname}" -- smbclient -d5 -N -L 127.0.0.1
+	echo -n "---- END DEBUG -----"
+	_error "Error: samba ad did not become reachable"
+fi
```
anoopcs9, line 57: I hope this additional change is not intended with this PR. If yes, are you planning to make such a change in _tests/test-deploy-ad-member.sh_?

anoopcs9 (changes requested):

tests/test-deploy-ad-server.sh
phlogistonjohn: It's a hack. When I run the script locally it passes. I have a personal convention of putting "xxx" in the commit message if its temporary/incomplete like some people put "wip".

images/server/Dockerfile.fedora
phlogistonjohn: Sure, but I want to keep things moving forward in a step-by-step manner. This just codifies what we already had.

images/server/Dockerfile.fedora
phlogistonjohn: Right now, I would prefer to stick with the hash. It helps me know exactly what we are taking into the container. I'm not against switching it, but I'd like to have something in place wrt samba-in-kubernetes/sambacc/issues/20  first

phlogistonjohn:
@anoopcs9 did you take a look at the failure itself, is it something familiar to you? Is it related to the recent cve fixes in samba?

images/server/Dockerfile.fedora
anoopcs9: Ok.

anoopcs9:
> @anoopcs9 did you take a look at the failure itself, is it something familiar to you? Is it related to the recent cve fixes in samba?

Ah.. I think you need to tag Fedora image in _images/ad-server/Containerfile_ too. Because it fails due to 4.15 version of Samba coming via Fedora 35 repositories.

tests/test-deploy-ad-server.sh
anoopcs9: I guessed right.

phlogistonjohn:
> > @anoopcs9 did you take a look at the failure itself, is it something familiar to you? Is it related to the recent cve fixes in samba?
> 
> Ah.. I think you need to tag Fedora image in _images/ad-server/Containerfile_ too. Because it fails due to 4.15 version of Samba coming via Fedora 35 repositories.

:facepalm: of course. thanks, will do!

anoopcs9 (changes requested):
Given that it passes all tests now, can we takeout the extra debug patch?

phlogistonjohn:
Debug removed. PTAL when you have a moment.

anoopcs9 (approved):

PR #57, opened 2021-12-16
server: match recent changes to sambacc packaging approach

phlogistonjohn:
sambacc now outputs wheels versioned based on the vcs tags and
vcs hash etc. as needed. It also now puts builds in a specific
"dist dir" under /srv when requested. This change makes the
samba-container match the new sambacc behavior.

### DIFF
--- a/images/server/Dockerfile.fedora
+++ b/images/server/Dockerfile.fedora
@@ -1,16 +1,16 @@
 FROM quay.io/samba.org/sambacc:latest AS builder
-ARG SAMBACC_VER=fd747527ed86d01821c4a3b72f1c4f893bb1ca36
+ARG SAMBACC_VER=master
 ARG SAMBACC_REPO=https://github.com/samba-in-kubernetes/sambacc
 
 # the changeset hash on the next line ensures we get a specifc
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
CI failure is expected until samba-in-kubernetes/sambacc#27 is merged. You can test this PR by hand locally by setting the build args SAMBACC_VER=jjm-package-setup  & SAMBACC_REPO=https://github.com/phlogistonjohn/sambacc when building the server image.

phlogistonjohn:
The sambacc changes are merged. So now this change passes CI.

anoopcs9 (commented):
Looks good. Don't we need similar changes in [Dockerfile.nightly](https://github.com/samba-in-kubernetes/samba-container/blob/master/images/server/Dockerfile.nightly)?

phlogistonjohn:
Yes indeed. Thanks for the reminder, I'll add to this PR.

In the longer term I think it would be a good idea to merge the two files and use build args to select between the different source package sets. That way we (I) avoid issues like this in the future.

anoopcs9 (approved):
lgtm, thanks.

raghavendra-talur (approved):

PR #58, opened 2021-12-31
Support configurable AD DC containers

phlogistonjohn:
At long last, I've finally got around to fixing this branch up. This now uses sambacc's `samba-dc-container` command to bootstrap the samba AD DC server. While it is not yet tested this gets us one step closer to configurable AD DC containers and, perhaps some time afterward, deployments with scale>1.


anoopcs9 (commented):
Current changes looks good.

I think we could also remove the following helper scripts as their functionality is available with `samba-dc-container` command from **_sambacc_**:
*  `provision.sh`
*  `populate.sh`
*  `run.sh`

phlogistonjohn:
@anoopcs9 I added a patch to remove the old files, as well as fixing the Makefile in the process.

Makefile
```diff
@@ -30,10 +30,8 @@ SERVER_SOURCES:=\
 	$(SERVER_DIR)/install-sambacc.sh
 AD_SERVER_SRC_FILE:=$(AD_SERVER_DIR)/Containerfile
 AD_SERVER_SOURCES:=\
-	$(AD_SERVER_DIR)/populate.sh \
-	$(AD_SERVER_DIR)/provision.sh \
-	$(AD_SERVER_DIR)/run.sh \
-	$(AD_SERVER_DIR)/install-packages.sh
+	$(AD_SERVER_DIR)/install-packages.sh \
+	$(AD_SERVER_DIR)/install-sambacc.sh
```
anoopcs9, line 34: Good catch.

anoopcs9 (approved):

PR #59, opened 2022-01-10
Consolidate samba server container build

phlogistonjohn:
Depends on: #57 

The first set of changes merges the previous fedora and nightly files into a single container file.

Remaining changes clean up a few old things.


anoopcs9 (commented):
Looks fine. But what about [smb.import.conf](https://github.com/samba-in-kubernetes/samba-container/blob/master/images/server/smb.import.conf)? Are we going to keep it under images/server?

phlogistonjohn:
> Looks fine. But what about [smb.import.conf](https://github.com/samba-in-kubernetes/samba-container/blob/master/images/server/smb.import.conf)? Are we going to keep it under images/server?

Looks like it is unused too. We can remove it.

images/server/install-packages.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -ex
+
+get_custom_repo() {
+    url="$1"
+    fname="$(basename "$url")"
+    dest="/etc/yum.repos.d/${fname}"
+    dnf install --setopt=install_weak_deps=False curl
```
phlogistonjohn, line 9: missing a `-y` here

raghavendra-talur (approved):

anoopcs9 (approved):
lgtm, thanks.

PR #60, opened 2022-01-11
Add shellcheck check rule

phlogistonjohn:
Based on PR: #59 

The samba-container project holds a fair number of shell scripts. Let's use the `shellcheck` static checker on them to avoid common errors.

The patches contain small fix-ups for the scripts, a new makefile rule, and a CI rule to execute the make rule.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -8,6 +8,27 @@ on:
 
 jobs:
 
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    # We need a newer version of shellcheck to avoid problems with the
+    # relative imports. Our scripts work on v0.7.2 and up but not the
+    # v0.7.0 preinstalled in the ubutnu image
+    - name: Update shellcheck
+      run: |
+        shellcheck_version="v0.8.0"
+        url="https://github.com/koalaman/shellcheck/releases/download/${shellcheck_version}/shellcheck-${shellcheck_version}.linux.x86_64.tar.xz"
+        curl -Lo /tmp/shellcheck.tar.xz "$url"
+        mkdir /tmp/shellcheck
+        tar -xf /tmp/shellcheck.tar.xz -C /tmp/shellcheck
+        mkdir -p ~/bin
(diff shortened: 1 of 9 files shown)


.github/workflows/container-image.yml
```diff
@@ -8,6 +8,27 @@ on:
...
+    # v0.7.0 preinstalled in the ubutnu image
+    - name: Update shellcheck
+      run: |
+        shellcheck_version="v0.8.0"
+        url="https://github.com/koalaman/shellcheck/releases/download/${shellcheck_version}/shellcheck-${shellcheck_version}.linux.x86_64.tar.xz"
+        curl -Lo /tmp/shellcheck.tar.xz "$url"
+        mkdir /tmp/shellcheck
+        tar -xf /tmp/shellcheck.tar.xz -C /tmp/shellcheck
+        mkdir -p ~/bin
+        install -m0755 /tmp/shellcheck/shellcheck-${shellcheck_version}/shellcheck  ~/bin/shellcheck
```
anoopcs9, line 26: Don't we have provision to update packages using default package manager? Or is it like the maximum available version on latest Ubuntu is not enough for us?

anoopcs9 (approved):

tests/test-deploy-ad-server.sh
```diff
@@ -76,15 +76,15 @@ echo >> "${TMPFILE}"
 FIRSTLINE="$(head -1 ./tests/files/coredns-snippet.template)"
 LASTLINE="    }"
 
-sed -i .backup -e "/$FIRSTLINE/,/$LASTLINE/d" ${TMPFILE}
+sed -i .backup -e "/$FIRSTLINE/,/$LASTLINE/d" "${TMPFILE}"
 
-cat tests/files/coredns-snippet.template \
-	| sed -e "s/AD_SERVER_IP/${AD_POD_IP}/" \
+sed -e "s/AD_SERVER_IP/${AD_POD_IP}/" \
+	< tests/files/coredns-snippet.template \
```
raghavendra-talur, line 82: I don't see a need for the "<" but I guess it does not do anything bad.

raghavendra-talur (approved):

PR #61, opened 2022-01-13
Run push jobs only within repository and not on forks

anoopcs9:
When someone forks the repository, they automatically get the GitHub actions workflow which would then run the push job without proper credentials to push images to quay.io/samba.org registry. Therefore we limit it by checking whether we are within original repository or not.


phlogistonjohn (approved):
Good catch, I had no idea that was happening. LGTM.

PR #64, opened 2022-01-31
server: allow post-install on image to fit OpenShift

synarete:
Some K8S environments (OpenShift) encorage not to run containers with UID=0,
though they do allow GID=0. Modify the relevant parts of root file-system
accordingly so that smbd will not get -EPERM when it is trying to modify
the local fs.

Signed-off-by: Shachar Sharon <ssharon@redhat.com>

### DIFF
--- a/images/server/Containerfile
+++ b/images/server/Containerfile
@@ -26,6 +26,8 @@ COPY install-sambacc.sh /usr/local/bin/install-sambacc.sh
 RUN /usr/local/bin/install-sambacc.sh \
     "/tmp" "/usr/local/share/sambacc/examples/minimal.json"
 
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
@synarete should we continue to leave this one open?  IIRC we decided to go with an approach more tied to the operator, right?  I think we can close this if so. Even after closing it we'll still be able to refer to it if needed.

Let me know what you think, thanks!

synarete:
> @synarete should we continue to leave this one open? IIRC we decided to go with an approach more tied to the operator, right? I think we can close this if so. Even after closing it we'll still be able to refer to it if needed.
> 
> Let me know what you think, thanks!

@phlogistonjohn No need for this one with latest samba-operator code. You may close it.

PR #66, opened 2022-02-16
workflows: consolidate build, test, and push actions

phlogistonjohn:
Consolidate all of our workflows into one yaml file. This means that the
"nightly" actions that build and test those images will be run on PRs.
Add the daily scheduled action that will test *and* push for all image
types not just the nightly rpms image.

Fixes: #65

PR #67, opened 2022-02-17
Build/Test/Push AD container images based on nightly Samba RPMs

anoopcs9:
This is something I had running under my fork for sometime and now rebased on top of https://github.com/samba-in-kubernetes/samba-container/pull/66.

@phlogistonjohn let me know your thoughts..

### DIFF
--- a/images/ad-server/Containerfile
+++ b/images/ad-server/Containerfile
@@ -1,7 +1,11 @@
 FROM registry.fedoraproject.org/fedora:34
 
-RUN dnf install -y samba-dc procps-ng /usr/bin/smbclient
-RUN rm -f /etc/samba/smb.conf
+ARG INSTALL_PACKAGES_FROM=default
+ARG SAMBA_SPECIFICS=
+
+COPY install-packages.sh /usr/local/bin/install-packages.sh
+RUN /usr/local/bin/install-packages.sh "$INSTALL_PACKAGES_FROM"
+
 COPY provision.sh /usr/local/bin/provision.sh
 COPY run.sh /usr/local/bin/run.sh
 COPY populate.sh /usr/local/bin/populate.sh
--- a/tests/test-deploy-ad-member.sh
+++ b/tests/test-deploy-ad-member.sh
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
 
 SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+IMG_TAG=${IMG_TAG:-"latest"}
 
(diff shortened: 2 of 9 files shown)


phlogistonjohn (changes requested):
First patch looks fine (changing the rpm repo location).

The next patch feels like it's doing to much in one patch and while I get the gist I have a difficult time following it.  Can I suggest you break it up into around 3 patches: one for the build changes, one for the test changes and one for the CI/infra changes? Thanks.

anoopcs9:
> First patch looks fine (changing the rpm repo location).
> 
> The next patch feels like it's doing to much in one patch and while I get the gist I have a difficult time following it. Can I suggest you break it up into around 3 patches: one for the build changes, one for the test changes and one for the CI/infra changes? Thanks.

Done.

images/ad-server/Containerfile
```diff
@@ -1,7 +1,10 @@
 FROM registry.fedoraproject.org/fedora:34
 
-RUN dnf install -y samba-dc procps-ng /usr/bin/smbclient
-RUN rm -f /etc/samba/smb.conf
+ARG INSTALL_PACKAGES_FROM=default
```
phlogistonjohn, line 3: Even if we don't do much with it, we should probably also have a "SAMBA_SPECIFICS" arg here like we do in the server image as a "breadcrumb"

tests/test-deploy-ad-member.sh
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env bash
 
 SCRIPT_DIR="$(cd "$(dirname "${0}")" && pwd)"
+IMG_TAG=${IMG_TAG:-"latest"}
 
 source "${SCRIPT_DIR}/common.sh"
 
 echo "Creating ad member pod..."
-ERROR_MSG=$(kubectl create -f "${MEMBER_POD_YAML}" 2>&1 1>/dev/null)
+ERROR_MSG=$(IMG_TAG=${IMG_TAG} envsubst < "${MEMBER_POD_YAML}" | kubectl create -f - 2>&1 1>/dev/null)
```
phlogistonjohn, line 9: Not my ideal form of templating, but I think its good enough for now.

phlogistonjohn (changes requested):
Only thing I needs doing is the SAMBA_SPECIFICS remark. Everything else is just commentary. Once that's done we can probably merge.

images/ad-server/Containerfile
anoopcs9: Fine.

phlogistonjohn (approved):

PR #69, opened 2022-03-02
makefile: add a rule to remove build files

phlogistonjohn:
Sometimes you don't want to skip building stuff. Add a simple 'clean'
rule that removes the "buildfiles".

Signed-off-by: John Mulligan <jmulligan@redhat.com>


anoopcs9 (approved):
clean change :-)

lgtm, thanks.

PR #71, opened 2022-05-16
Install sambacc from rpm

phlogistonjohn:
Depends on: samba-in-kubernetes/sambacc#40

Now that sambacc's build image can generate an RPM, use that to install sambacc. I retained the ability to install from a python wheel for now in case we find something strange with the RPM and need to revert. In the not too distant future, if all is fine with the rpm, we can probably drop support from installing from the wheel (unless we decide we want to support non-rpm distros?)

### DIFF
--- a/images/ad-server/install-sambacc.sh
+++ b/images/ad-server/install-sambacc.sh
@@ -3,18 +3,46 @@
 set -ex
 
 wheeldir="$1"
-container_json_file="$2"
-
-wheeldir=/tmp
-wheel="$(find "${wheeldir}" -type f -name 'sambacc-*.whl')" \
-
-if ! [ "$(echo "$wheel" | wc -l)" = 1 ]; then
-    echo "more than one wheel file found"
-    exit 1
+if ! [ -d "${wheeldir}" ]; then
+    echo "no directory: ${wheeldir}"
+    exit 2
 fi
 
-pip install "$wheel"
-rm -f "$wheel"
+mapfile -d '' wheels < \
+    <(find "${wheeldir}" -type f -name 'sambacc-*.whl' -print0)
+mapfile -d '' rpmfiles < \
+    <(find "${wheeldir}" -type f -name '*sambacc-*.noarch.rpm' -print0)
+
+
+if [ "${#wheels[@]}" -gt 1 ]; then
+    echo "more than one wheel file found"
+    exit 1
+elif [ "${#wheels[@]}" -eq 1 ]; then
+    action=install-wheel
+fi
+
+if [ "${#rpmfiles[@]}" -gt 1 ]; then
+    echo "more than one rpm file found"
+    exit 1
+elif [ "${#rpmfiles[@]}" -eq 1 ]; then
+    action=install-rpm
+fi
+
+case $action in
+    install-wheel)
+        pip install "${wheels[0]}"
+        container_json_file="/usr/local/share/sambacc/examples/addc.json"
+    ;;
+    install-rpm)
+        dnf install -y "${rpmfiles[0]}"
+        dnf clean all
+        container_json_file="/usr/share/sambacc/examples/addc.json"
+    ;;
+    *)
+        echo "no install package(s) found"
+        exit 1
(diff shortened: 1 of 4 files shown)


images/ad-server/install-sambacc.sh
```diff
@@ -3,18 +3,44 @@
...
 
 wheeldir=/tmp
-wheel="$(find "${wheeldir}" -type f -name 'sambacc-*.whl')" \
+find "${wheeldir}"
+wheel="$(find "${wheeldir}" -type f -name 'sambacc-*.whl')"
+rpmfile="$(find "${wheeldir}" -type f -name '*sambacc-*.noarch.rpm')"
 
-if ! [ "$(echo "$wheel" | wc -l)" = 1 ]; then
-    echo "more than one wheel file found"
+rpmcount="$(echo "$rpmfile" | grep -c rpm)"
```
synarete, line 12: When I tried to build with those changes on my machine (Fedora35), it fails on this line. The exit code of `grep` when there is no rpm is 1, which in turn causes this shell script to fail due to the `set -ex` at line 3.

images/server/install-sambacc.sh
```diff
@@ -3,18 +3,44 @@
...
 
 wheeldir=/tmp
-wheel="$(find "${wheeldir}" -type f -name 'sambacc-*.whl')" \
+find "${wheeldir}"
+wheel="$(find "${wheeldir}" -type f -name 'sambacc-*.whl')"
+rpmfile="$(find "${wheeldir}" -type f -name '*sambacc-*.noarch.rpm')"
 
-if ! [ "$(echo "$wheel" | wc -l)" = 1 ]; then
-    echo "more than one wheel file found"
+rpmcount="$(echo "$rpmfile" | grep -c rpm)"
```
synarete, line 12: Same comment as above, regrading grep exit with 1 which causes the entire script to fail due to `set -ex` at line 3.

synarete (commented):
Failed to build on my machine (F35)

images/ad-server/install-sambacc.sh
synarete: You are correct: as soon as I deleted my older sambacc image, everything works fine. At least it helped catch this one!
Other than that, LGTM.

images/ad-server/install-sambacc.sh
```diff
@@ -3,18 +3,46 @@
...
+elif [ "${#wheels[@]}" -eq 1 ]; then
+    action=install-wheel
+fi
+
+if [ "${#rpmfiles[@]}" -gt 1 ]; then
+    echo "more than one rpm file found"
+    exit 1
+elif [ "${#rpmfiles[@]}" -eq 1 ]; then
+    action=install-rpm
 fi
```
anoopcs9, line 29: Can you explain the logic here? Why do we insist on having just a single rpm? Under what circumstances we expect more than one?

images/server/install-sambacc.sh
```diff
@@ -3,18 +3,46 @@
...
+elif [ "${#wheels[@]}" -eq 1 ]; then
+    action=install-wheel
+fi
+
+if [ "${#rpmfiles[@]}" -gt 1 ]; then
+    echo "more than one rpm file found"
+    exit 1
+elif [ "${#rpmfiles[@]}" -eq 1 ]; then
+    action=install-rpm
 fi
```
anoopcs9, line 29: Same comment as above.

images/ad-server/install-sambacc.sh
phlogistonjohn: The build container should only generate exactly one RPM (recent versions) and exactly one wheel (traditionally) but we also dont know for sure that the input is the result of the sambacc build container. Ensuring that we only have 1 RPM or wheel in the source dir is a simplistic way to try and double check that our input dir is what we expect.  If you mounted a dir from your host that had multiple extra files in it would be one way to trigger the condition where there are multiple RPMs or wheels.

anoopcs9 (approved):
lgtm.

PR #72, opened 2022-05-18
Install python3-pyxattr package on server image

spuiuk:
The new permissions feature to sambacc required the python xattr module.

Signed-off-by: Sachin Prabhu <sprabhu@redhat.com>

### DIFF
--- a/images/ad-server/install-packages.sh
+++ b/images/ad-server/install-packages.sh
@@ -25,6 +25,7 @@ dnf install --setopt=install_weak_deps=False -y \
     python-pip \
     python3-jsonschema \
     python3-samba \
+    python3-pyxattr \
     samba-dc \
     procps-ng \
     /usr/bin/smbclient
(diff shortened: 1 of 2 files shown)


anoopcs9:
With https://github.com/samba-in-kubernetes/sambacc/pull/39 merged I think python-pyxattr installation is no longer required.

spuiuk:
You are right. We will not need this patch request once PR https://github.com/samba-in-kubernetes/sambacc/pull/40 is accepted as the sambacc rpm will pull in the required dependency. But let's just leave this PR until the sambacc rpms PR is merged.

Just for clarification, the Sambacc PR 39 skips xattr tests if the test environment doesn't support xattrs. However we will still need xattr support for normal operations(ie. for Permissions change feature).

phlogistonjohn (approved):
LGTM, you can take it out of draft and we can merge at any time IMO.

PR #74, opened 2022-05-31
Specify a version suffix to pin samba versions

phlogistonjohn:
This applies to the server and ad-server container images.

The planned release branch will be able to set the SAMBA_VERSION_SUFFIX build arg to a specific version.

### DIFF
--- a/images/ad-server/Containerfile
+++ b/images/ad-server/Containerfile
@@ -11,12 +11,15 @@ RUN SAMBACC_DISTNAME=latest \
 
 FROM registry.fedoraproject.org/fedora:35
 ARG INSTALL_PACKAGES_FROM=default
+ARG SAMBA_VERSION_SUFFIX=""
 ARG SAMBA_SPECIFICS=
 
 MAINTAINER John Mulligan <jmulligan@redhat.com>
 
 COPY install-packages.sh /usr/local/bin/install-packages.sh
-RUN /usr/local/bin/install-packages.sh "$INSTALL_PACKAGES_FROM"
+RUN /usr/local/bin/install-packages.sh \
+    "${INSTALL_PACKAGES_FROM}" \
+    "${SAMBA_VERSION_SUFFIX}"
 
 COPY --from=builder \
     /srv/dist/latest \
--- a/images/server/Containerfile
+++ b/images/server/Containerfile
@@ -11,13 +11,16 @@ RUN SAMBACC_DISTNAME=latest \
 
 FROM registry.fedoraproject.org/fedora:35
 ARG INSTALL_PACKAGES_FROM=default
+ARG SAMBA_VERSION_SUFFIX=""
 ARG SAMBA_SPECIFICS=daemon_cli_debug_output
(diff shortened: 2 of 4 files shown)


anoopcs9 (commented):
I hope `SAMBA_VERSION_SUFFIX` is expected to be in the format _X.Y.Z_. In that case successful installation of packages would then require a hyphen(`-`) between package name and `SAMBA_VERSION_SUFFIX`. For example:

```
"samba-${samba_version_suffix}"
"samba-dc-${samba_version_suffix}"

File: images/ad-server/Containerfile [hunk 1]
```diff
@@ -11,12 +11,15 @@ RUN SAMBACC_DISTNAME=latest \
...
+ARG SAMBA_VERSION_SUFFIX=""
 ARG SAMBA_SPECIFICS=
 
 MAINTAINER John Mulligan <jmulligan@redhat.com>
 
 COPY install-packages.sh /usr/local/bin/install-packages.sh
-RUN /usr/local/bin/install-packages.sh "$INSTALL_PACKAGES_FROM"
+RUN /usr/local/bin/install-packages.sh \
+    "$INSTALL_PACKAGES_FROM" \
+    "${SAMBA_VERSION_SUFFIX}"
```
synarete, line 22: Style: either use curly braces in both variables substitutions, or in none.

images/server/Containerfile
```diff
@@ -11,13 +11,16 @@ RUN SAMBACC_DISTNAME=latest \
...
 ARG SAMBA_SPECIFICS=daemon_cli_debug_output
 
 MAINTAINER John Mulligan <jmulligan@redhat.com>
 
 COPY smb.conf /etc/samba/smb.conf
 COPY install-packages.sh /usr/local/bin/install-packages.sh
-RUN /usr/local/bin/install-packages.sh "$INSTALL_PACKAGES_FROM"
+RUN /usr/local/bin/install-packages.sh \
+    "$INSTALL_PACKAGES_FROM" \
+    "${SAMBA_VERSION_SUFFIX}"
```
synarete, line 23: Style: either use curly braces in both variables substitutions, or in none.

synarete (commented):
Other then style comments, LGTM

phlogistonjohn:
> I hope `SAMBA_VERSION_SUFFIX` is expected to be in the format _X.Y.Z_. In that case successful installation of packages would then require a hyphen(`-`) between package name and `SAMBA_VERSION_SUFFIX`. For example:
> 
> ```
> "samba-${samba_version_suffix}"
> "samba-dc-${samba_version_suffix}"
> ```

I had called the variable "samba_version_suffix" because you must supply the dash. It sounds to me that you don't like this but I'm not 100% sure about that. I had decided to require the dash to be part of the var in order to make the script extra simple. Please let me know if you object to this approach or not, thanks!

images/server/Containerfile
phlogistonjohn: fixed

PR #75, opened 2022-06-22
Utilize build args to install from custom repo

anoopcs9:
Even though we had **INSTALL_CUSTOM_REPO** variable consumed inside _install-packages.sh_ script it was never exposed to be considered as an input. Therefore we now utilize `--build-arg` to provide valid repository URLs for installing packages.


phlogistonjohn (approved):
LGTM.

phlogistonjohn:
I'll leave it unmerged so others can provide feedback if desired, but I'll merge in a few hours if no one else chimes in.

Issue #73, opened 2022-05-18
ci: reuse images across build steps


phlogistonjohn:
I think it should be possible to share the images built by the dedicated build steps with the steps that execute the test scripts in the github ci.  At the very least I'd like to investigate if it is possible and know once and for all.

The goal would be to speed up the CI and avoid redundant build steps.

anoopcs9:
Possible lengthy approach using [upload-artifact](https://github.com/actions/upload-artifact) and [download-artifact](https://github.com/actions/download-artifact):
```diff --git a/.github/workflows/container-image.yml b/.github/workflows/container-image.yml
index cfc1ae5..8ce78b2 100644
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -38,7 +38,15 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Build the server image
-      run: make build-server
+      run: |
+        make build-server
+        podman save samba-container:latest > /tmp/server-image.tar
+    - name: Save image
+      uses: actions/upload-artifact@v3
+      with:
+        name: server-image-artifact
+        path: /tmp/server-image.tar
+        retention-days: 1
 
   build-ad-server:
     runs-on: ubuntu-latest
@@ -47,7 +55,15 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Build the ad server image
-      run: make build-ad-server
+      run: |
+        make build-ad-server
+        podman save samba-ad-container:latest > /tmp/ad-server-image.tar
+    - name: Save image
+      uses: actions/upload-artifact@v3
+      with:
+        name: ad-server-image-artifact
+        path: /tmp/ad-server-image.tar
+        retention-days: 1
 
   build-client:
     runs-on: ubuntu-latest
@@ -62,9 +78,19 @@ jobs:
     #needs: build-server
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - name: Retrieve image
+      uses: actions/download-artifact@v3
+      with:
+        name: server-image-artifact
+        path: /tmp
+    - name: Load image
+      run: podman load --input /tmp/server-image.tar
     - name: Test the server image
-      run: make test-server
+      uses: actions/checkout@v2
+      run: |
+        > .build-server
+        make test-server
+
 
   # Reminder: the nightly-server images consume nightly samba rpm builds
   # it is not *just* an image that gets built nightly
[...]

PR #76, opened 2022-07-21
workflows: Share images between jobs

anoopcs9:
This is an attempt to evaluate and measure the effort to reuse images built in earlier steps while running actual tests so as to improve overall test time. Here we make use of [upload-artifact](https://github.com/actions/upload-artifact) in build step and [download-artifact](https://github.com/actions/download-artifact) in actual test run step.

fixes: #73

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -8,6 +8,9 @@ on:
   schedule:
     - cron: '0 2 * * *'
 
+env:
+  CONTAINER_CMD: docker
+
 jobs:
 
   check:
@@ -39,6 +42,11 @@ jobs:
     - uses: actions/checkout@v2
     - name: Build the server image
       run: make build-server
+    - name: Upload server image
+      uses: ishworkh/docker-image-artifact-upload@v1
+      with:
+        image: "samba-container:latest"
+        retention_days: 1
 
   build-ad-server:
     runs-on: ubuntu-latest
@@ -48,6 +56,11 @@ jobs:
     - uses: actions/checkout@v2
     - name: Build the ad server image
       run: make build-ad-server
+    - name: Upload ad server image
+      uses: ishworkh/docker-image-artifact-upload@v1
+      with:
+        image: "samba-ad-container:latest"
+        retention_days: 1
 
   build-client:
     runs-on: ubuntu-latest
@@ -59,12 +72,16 @@ jobs:
       run: make build-client
 
   test-server:
-    #needs: build-server
+    needs: build-server
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - name: Download server image
+      uses: ishworkh/docker-image-artifact-download@v1
+      with:
+        image: "samba-container:latest"
     - name: Test the server image
-      run: make test-server
+      run: tests/test-samba-container.sh
 
   # Reminder: the nightly-server images consume nightly samba rpm builds
   # it is not *just* an image that gets built nightly
@@ -76,6 +93,11 @@ jobs:
     - uses: actions/checkout@v2
     - name: Build the nightly server image
       run: make build-nightly-server
+    - name: Upload nightly server image
+      uses: ishworkh/docker-image-artifact-upload@v1
+      with:
+        image: "samba-container:nightly"
+        retention_days: 1
 
   build-nightly-ad-server:
     runs-on: ubuntu-latest
@@ -85,16 +107,28 @@ jobs:
     - uses: actions/checkout@v2
     - name: Build the nightly ad server image
       run: make build-nightly-ad-server
+    - name: Upload nightly ad server image
+      uses: ishworkh/docker-image-artifact-upload@v1
+      with:
+        image: "samba-ad-container:nightly"
+        retention_days: 1
 
   test-nightly-server:
-    #needs: build-nightly-server
+    needs: build-nightly-server
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
+    - name: Download nightly server image
+      uses: ishworkh/docker-image-artifact-download@v1
+      with:
+        image: "samba-container:nightly"
     - name: Test the nightly server image
-      run: make test-nightly-server
+      run: LOCAL_TAG=samba-container:nightly tests/test-samba-container.sh
 
   test-ad-server-kubernetes:
+    needs:
+      - build-ad-server
+      - build-server
     #runs-on: ubuntu-latest
     # need to explicitly use 20.04 to avoid problems with jq...
     runs-on: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


anoopcs9:
Interesting..CI runs aren't getting triggered?

phlogistonjohn:
Maybe start with a smaller delta between the original versions of the yamls and make smaller steps to see what is making github unhappy?  Or maybe it's just a fluke :shrug:

anoopcs9:
> Maybe start with a smaller delta between the original versions of the yamls and make smaller steps to see what is making github unhappy? Or maybe it's just a fluke shrug

Thanks for the hint. I force pushed with a smaller hunk and figured out that _actions/checkout@v2_ was misplaced.

anoopcs9:
I ran CI multiple times and compared the average with previous runs. Maximum improvement w.r.t overall time taken was around 20%, i.e from ~10m+ to ~8m+.

phlogistonjohn:
> I ran CI multiple times and compared the average with previous runs. Maximum improvement w.r.t overall time taken was around 20%, i.e from ~10m+ to ~8m+.

That's a pleasant surprise, I guess since the object store and the runners are all in azure the overhead of communicating with the object store is less than say, me at my house. :-)

Doing this also has the interesting side-effect that if you needed to you could pull the artifact down locally and use it for local testsing, for example if the test was failing and you couldn't tell why. You'd even be sure that you were consuming the exact image the ci test was.

Given that, I'm good to go with the approach. I'll actually start reviewing the code.

.github/workflows/container-image.yml
```diff
@@ -38,7 +38,15 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Build the server image
-      run: make build-server
+      run: |
+        make build-server
```
phlogistonjohn, line 42: Is there are reason this isn't two run steps? LIke:
```
- name: Build the server image
  make build-server
- name: Export the server image
  podman save samba-container:latest > /tmp/server-image.tar
```

.github/workflows/container-image.yml
```diff
@@ -59,12 +75,21 @@ jobs:
...
     - uses: actions/checkout@v2
+    - name: Retrieve image
+      uses: actions/download-artifact@v3
+      with:
+        name: server-image-artifact
+        path: /tmp
+    - name: Load image
+      run: |
+        docker load --input /tmp/server-image.tar
+        docker tag localhost/samba-container:latest samba-container:latest
```
phlogistonjohn, line 90: Why is the `docker tag ....` line necessary? Doesn't docker (and podman) treat `localhost/blah:foo == blah:foo` ?  If you tried it and it doesn't work, no worries. I'm just curious because I thought it did.

.github/workflows/container-image.yml
```diff
@@ -117,6 +184,9 @@ jobs:
   test-nightly-ad-server-kubernetes:
       #runs-on: ubuntu-latest
       # need to explicitly use 20.04 to avoid problems with jq...
+      needs:
+        - build-nightly-ad-server
+        - build-nightly-server
```
phlogistonjohn, line 189: Putting the lines here splits the comment above from the line it's remarking upon (the runs-on line below).
This probably needs to be shifted up or down to avoid this splitting.

phlogistonjohn (changes requested):
Only one must fix: the comment/runs-on split.

----

Everything else seems reasonable but I do have a few additional remarks.
One thing I couldn't avoid thinking as I read it: wouldn't it be nice if some of the details were hidden by a higher level action like:
```
- name: Save image
  uses: imaginary/image-saver@v1
  image: samba-container:latest
  destination: server-image-artifact

OR

- name: Load image
  uses: imaginary/image-saver@v1
  image: samba-container:latest
  source: server-image-artifact
```

I have no idea if this exists already, but it may be worth looking if you have not already. If it doesn't exist maybe it would be cool to make it? I have no idea how to make an action but it may be worthwhile (for the future of course )

.github/workflows/container-image.yml
anoopcs9: Fixed.

.github/workflows/container-image.yml
anoopcs9: Make sense. Updated in every other occurrence.

.github/workflows/container-image.yml
anoopcs9: Even I thought so but its not the case it seems.

anoopcs9:
> One thing I couldn't avoid thinking as I read it: wouldn't it be nice if some of the details were hidden by a higher level action like:
> 
> ```
> - name: Save image
>   uses: imaginary/image-saver@v1
>   image: samba-container:latest
>   destination: server-image-artifact
> 
> OR
> 
> - name: Load image
>   uses: imaginary/image-saver@v1
>   image: samba-container:latest
>   source: server-image-artifact
> ```
> 
> I have no idea if this exists already, but it may be worth looking if you have not already. If it doesn't exist maybe it would be cool to make it? I have no idea how to make an action but it may be worthwhile (for the future of course )

A quick search over marketplace listed the following actions which matches the above expectation:
* [docker-image-artifact-upload](https://github.com/marketplace/actions/docker-image-artifact-upload)
* [docker-image-artifact-download](https://github.com/marketplace/actions/docker-image-artifact-download)

But it lacks `retention-days` support for which a [PR](https://github.com/ishworkh/docker-image-artifact-upload/pull/6) is already raised.

Shall we give it a try?

phlogistonjohn:
> 
> But it lacks `retention-days` support for which a [PR](https://github.com/ishworkh/docker-image-artifact-upload/pull/6) is already raised.
> 
> Shall we give it a try?

LGTM, so why not.

anoopcs9:
> > But it lacks `retention-days` support for which a [PR](https://github.com/ishworkh/docker-image-artifact-upload/pull/6) is already raised.
> > Shall we give it a try?
> 
> LGTM, so why not.

I've switched to `docker` instead of `podman` globally in the workflow to satisfy the requirement for custom action used. Was there an issue with `docker` pushing images to Quay repository?

phlogistonjohn:
> > > But it lacks `retention-days` support for which a [PR](https://github.com/ishworkh/docker-image-artifact-upload/pull/6) is already raised.
> > > Shall we give it a try?
> > 
> > 
> > LGTM, so why not.
> 
> I've switched to `docker` instead of `podman` globally in the workflow to satisfy the requirement for custom action used. Was there an issue with `docker` pushing images to Quay repository?

I don't think so. I think we just wanted to use podman where possible since we like it.

I have no issue with this change but it might be worth reporting an issue to that repo to request an option to use podman? Then perhaps we could use it in the future once again.

phlogistonjohn (approved):
LGTM

phlogistonjohn:
Since this is a lot of lines subtly changed I'd like a 2nd review on this. @synarete do you have a moment to review this?

synarete:
> 

@phlogistonjohn  Sure, I will take a look in 5 minutes

synarete (approved):
LGTM

@anoopcs9  One minor question: I do not see anything which have `needs: build-client`. Shouldn't we add such `need`, at least to  `push` ?

anoopcs9:
> LGTM
> 
> @anoopcs9 One minor question: I do not see anything which have `needs: build-client`. Shouldn't we add such `need`, at least to `push` ?

I haven't taken **push** job into consideration since that's something which happens after PR is approved(and merged) or as part of scheduled run. But if we decide to add `needs` for `build-client` and download before pushing client image then we could add similar steps for downloading other images too in `push` job. Nevertheless, I am very much open to such an additional change.

@phlogistonjohn What do you think?

anoopcs9:
> I have no issue with this change but it might be worth reporting an issue to that repo to request an option to use podman? Then perhaps we could use it in the future once again.

I have reported https://github.com/ishworkh/docker-image-artifact/issues/3. Let's see..

phlogistonjohn (approved):
LGTM

synarete (approved):
LGTM

PR #79, opened 2022-09-20
images/server: Use new CTDB 'leader' admin sub-command

anoopcs9:
Use the detection mechanism from sambacc(https://github.com/samba-in-kubernetes/sambacc/pull/50) via `SAMBA_SPECIFICS` env variable to indicate the presence of new/replaced administrative sub-command 'leader' for CTDB.


phlogistonjohn (approved):
LGTM

PR #83, opened 2022-11-10
images/client: add 'samba-test'

synarete:
Allow using the samba-client image for testing, via the 'smbtorture' utility.

Signed-off-by: Shachar Sharon <ssharon@redhat.com>

### DIFF
--- a/images/client/Dockerfile
+++ b/images/client/Dockerfile
@@ -4,4 +4,4 @@ FROM quay.io/centos/centos:stream8
 
 MAINTAINER Michael Adam <obnox@samba.org>
 
-RUN dnf -y install samba-client
+RUN dnf -y install samba-client samba-test


synarete:
> Thinking out loud: would this be better in a separate image? While it's nice not to have to manage lots of images we're asking every consumer of a samba client image to also carry around stuff that's only useful to those testing samba. Basically, is the samba client image meant more for our own testing or do we want it to be for a broader audience? Note that neither choice is set in stone - we can decide that for now, the image is more for us, but if more consumers of the image come along we can split things up in the future. -- Food for thought, especially while the PR is still draft :-)

I follow the same line of thought, which is why I turned this PR into draft. Still not sure which option is the best.

phlogistonjohn:
Sounds good to me, having a specific "toolbox" image would be a clean approach.

synarete:
> Sounds good to me, having a specific "toolbox" image would be a clean approach.

See: https://github.com/samba-in-kubernetes/samba-container/pull/84

obnoxxx (changes requested):
please rebase and resolve conflicts.

did you mean to add smbtorture instead of samba-test?

synarete:
> please rebase and resolve conflicts.
> 
> did you mean to add smbtorture instead of samba-test?

@obnoxxx  'samba-test' is the name of the rpm package which contains smbtorture. In all odds, this PR has changed to draft, as we decided to use a different image for toolbox (which will be based of samba-client image).

obnoxxx:
> > please rebase and resolve conflicts.
> > did you mean to add smbtorture instead of samba-test?
> 
> @obnoxxx 'samba-test' is the name of the rpm package which contains smbtorture. In all odds, this PR has changed to draft, as we decided to use a different image for toolbox (which will be based of samba-client image).

@synarete , thanks for the explanation. it all looks good, and the toolbox PR #84 is merged now

PR #84, opened 2022-11-13
samba toolbox

synarete:
Create dedicate image for developer's or power-user admin and testing operations. Similar to rook's toolbox.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -71,6 +71,15 @@ jobs:
     - name: Build the client image
       run: make build-client
 
+  build-toolbox:
+    runs-on: ubuntu-latest
+    env:
+      BUILDAH_FORMAT: oci
+    steps:
+    - uses: actions/checkout@v3
+    - name: Build the client image
+      run: make build-toolbox
+
   test-server:
     needs: build-server
     runs-on: ubuntu-latest
@@ -185,6 +194,7 @@ jobs:
     # verify it passes the test jobs first
     needs:
       - build-client
+      - build-toolbox
       - test-server
       - test-nightly-server
       - test-ad-server-kubernetes
@@ -201,6 +211,8 @@ jobs:
         run: make push-ad-server
       - name: push client image
         run: make push-client
+      - name: push toolbox image
+        run: make push-toolbox
       - name: push nightly server image
         run: make push-nightly-server
       - name: push nightly ad server image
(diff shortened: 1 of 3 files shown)


phlogistonjohn (approved):
LGTM.

I'll leave it open a couple of days for others to take a look. Feel free to merge by Thursday if you don't get any other feedback.

anoopcs9 (commented):
Should we then keep `samba-client` image as a separate one? How about renaming the current `samba-client` container?

synarete:
> Should we then keep `samba-client` image as a separate one? How about renaming the current `samba-client` container?

This was an option we considered, but decided it would be better to leave `samba-client` a minimal image and have separate `samba-toolbox`, which most likely will add more and more debug packages over time (as needed).

anoopcs9:
> > Should we then keep `samba-client` image as a separate one? How about renaming the current `samba-client` container?
> 
> This was an option we considered, but decided it would be better to leave `samba-client` a minimal image and have separate `samba-toolbox`, which most likely will add more and more debug packages over time (as needed).

Just now happened to notice the discussion from https://github.com/samba-in-kubernetes/samba-container/pull/83. Installing `samba-test` must be pulling more dependencies than `samba-client`. I liked the _toolbox_ suffix which can hold `samba-client` too. Considering the very low(or none) consumption of `samba-client` outside our testing infrastructure I am slightly more inclined to include `samba-test` and thereby rename the current image.

Sorry for the late noise.

synarete:
> > > Should we then keep `samba-client` image as a separate one? How about renaming the current `samba-client` container?
> > 
> > 
> > This was an option we considered, but decided it would be better to leave `samba-client` a minimal image and have separate `samba-toolbox`, which most likely will add more and more debug packages over time (as needed).
> 
> Just now happened to notice the discussion from #83. Installing `samba-test` must be pulling more dependencies than `samba-client`. I liked the _toolbox_ suffix which can hold `samba-client` too. Considering the very low(or none) consumption of `samba-client` outside our testing infrastructure I am slightly more inclined to include `samba-test` and thereby rename the current image.
> 
> Sorry for the late noise.

That is also a valid option (thou it would require also a change in places which currently use samba-client image).
@phlogistonjohn What do you think?

phlogistonjohn:
In quite a few cases I like the "fake it till you make it" approach unless there is a significant maintenance burden. In this cases "faking it" means we act like we have external users consuming samba-client. Adding `samba-test` or `samba-toolbox` means we don't disrupt our imaginary samba-client users with lots of extra stuff in the image. We could even choose to make samba-client a lower layer in the image - if the base distros align.  I'm not adamant about keeping samba-client but that's my current preference.

obnoxxx (approved):
since this has been agreed upon in principle in the discussions, I am approving. lgtm!

PR #92, opened 2023-01-16
workflows: add commitlint workflow

obnoxxx:
component definitions are basically taken from the existing git log.

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,0 +1,31 @@
+{
+  "extends": [
+    "@commitlint/config-conventional"
+  ],
+  "rules": {
+    "type-enum": [
+      2,
+      "always",
+      [
+        "server",
+        "ad-server",
+        "tests",
+        "images",
+        "docs",
+        "examples",
+        "workflows",
+        "kubernetes",
+        "hack",
+	"Makefile"
+	]
+    ],
+    "body-leading-blank": [
+      2,
+      "always"
+    ],
+    "footer-leading-blank": [
+      2,
+      "always"
+    ]
+  }
+}
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -1,0 +1,26 @@
+name: Commitlint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - release-*
+  pull_request:
+    branches:
+      - master
+      - release-*
+
+jobs:
+  lint:
+    runs-on: ubuntu-18.04
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: wagoid/commitlint-github-action@v2.0.3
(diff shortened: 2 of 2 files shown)


obnoxxx:
strange, I don't see why the new workflow file does not trigger a check run ... 🤷‍♂️

obnoxxx:
> strange, I don't see why the new workflow file does not trigger a check run ... 🤷‍♂️

now it starts and even succeeds! 😄

.commitlintrc.json
```diff
@@ -0,0 +1,31 @@
...
+        "server",
+        "ad-server",
+        "tests",
+        "images",
+        "docs",
+        "examples",
+        "workflows",
+        "kubernetes",
+        "hack",
+	"Makefile"
```
synarete, line 19: We user `makefile:` (lowercase) labels so far. Better keep it that way.

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,26 @@
+name: Commitlint
```
synarete, line 1: Unlike `rook`, in `samba-operator` we use single workflow-action yaml, which allows us to see all jobs at once (as DAG) and halt workflow as soon a a job fails. Better have this section part of `main.yaml` with proper `needs`.

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,26 @@
...
+      - master
+      - release-*
+  pull_request:
+    branches:
+      - master
+      - release-*
+
+jobs:
+  lint:
+    runs-on: ubuntu-18.04
```
synarete, line 16: Use `ubuntu-latest`

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,26 @@
...
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
```
synarete, line 20: Better use latest `actions/checkout@v3`

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          fetch-depth: 0
```
synarete, line 22: Why do you need this? AFAIK, it is not needed in this case.

.github/workflows/commitlint.yml
```diff
@@ -0,0 +1,26 @@
...
+      - uses: wagoid/commitlint-github-action@v2.0.3
```
synarete, line 23: Use latest version: `wagoid/commitlint-github-action@v5`

synarete (commented):
The type-enum looks ok, but see other comments in this PR

obnoxxx:
converted to draft becaus @phlogistonjohn suggested we solve the corresponding effort for samba-operator first : https://github.com/samba-in-kubernetes/samba-operator/pull/272 or https://github.com/samba-in-kubernetes/samba-operator/pull/274

anoopcs9:
Support for gitlint is already merged via #124.

Issue #78, opened 2022-08-31
Switch to `centos:stream9` base image


anoopcs9:
Use `quay.io/centos/centos:stream9` as the base image for building Samba server and AD server images.

anoopcs9:
`python-inotify_simple` is now [available](https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2022-c3b6e6b1f3) with EPEL 9 and an attempt to build sambacc was [successful](https://github.com/samba-in-kubernetes/sambacc/commit/1eff2a4c9e4a0a262cc7a35a54133aa20d58274e). But then while trying to build from samba-container repository I realized that we miss `samba-dc` altogether from RHEL 😕 . Still plain server images can be [built](https://github.com/anoopcs9/samba-container/commit/35428c6854da58a90b3780c92a45ed58bc6e710d).

ibotty:
Any news?

obnoxxx:
do we still want this, or rather close it, @phlogistonjohn , @anoopcs9 ?

anoopcs9:
> do we still want this, or rather close it, @phlogistonjohn , @anoopcs9 ?

This is an issue and can probably stay open.

anoopcs9:
_Update_:

Server images can be built with CentOS base.
AD Server images(with Samba packages from nightly RPM repository) can be built with CentOS base.

obnoxxx:
@ibotty  wrote: 

> Any news?


As @anoopcs9  has mentioned in the previous comment, I think this has long been done:

* ad-server: #142 
* server: #98 

closing this issue as completed.

PR #98, opened 2023-01-18
images switch server to centos:stream9 as base image

obnoxxx:
Fixes: #78

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ SERVER_DIR:=images/server
 AD_SERVER_DIR:=images/ad-server
 CLIENT_DIR:=images/client
 TOOLBOX_DIR:=images/toolbox
-SERVER_SRC_FILE:=$(SERVER_DIR)/Containerfile.fedora
+SERVER_SRC_FILE:=$(SERVER_DIR)/Containerfile.centos
 SERVER_SOURCES:=\
 	$(SERVER_DIR)/smb.conf \
 	$(SERVER_DIR)/install-packages.sh \
@@ -39,7 +39,7 @@ TOOLBOX_SRC_FILE:=$(TOOLBOX_DIR)/Containerfile.fedora
 TAG?=latest
 NIGHTLY_TAG?=nightly
 
-OS_NAME?= fedora
+OS_NAME?=centos
 
 
 _REPO_BASE?=quay.io//samba.org/
--- a/Makefile
+++ b/Makefile
@@ -37,26 +37,49 @@ CLIENT_SRC_FILE:=$(CLIENT_DIR)/Containerfile.fedora
 TOOLBOX_SRC_FILE:=$(TOOLBOX_DIR)/Containerfile.fedora
 
 TAG?=latest
-SERVER_NAME:=samba-container:$(TAG)
-NIGHTLY_SERVER_NAME:=samba-container:nightly
-AD_SERVER_NAME:=samba-ad-container:$(TAG)
-NIGHTLY_AD_SERVER_NAME:=samba-ad-container:nightly
-CLIENT_NAME:=samba-client-container:$(TAG)
-TOOLBOX_NAME:=samba-toolbox-container:$(TAG)
+NIGHTLY_TAG?=nightly
 
-SERVER_REPO_NAME:=quay.io/samba.org/samba-server:$(TAG)
-NIGHTLY_SERVER_REPO_NAME:=quay.io/samba.org/samba-server:nightly
-AD_SERVER_REPO_NAME:=quay.io/samba.org/samba-ad-server:$(TAG)
-NIGHTLY_AD_SERVER_REPO_NAME:=quay.io/samba.org/samba-ad-server:nightly
-CLIENT_REPO_NAME:=quay.io/samba.org/samba-client:$(TAG)
-TOOLBOX_REPO_NAME:=quay.io/samba.org/samba-toolbox:$(TAG)
+OS_NAME?= fedora
 
-BUILDFILE_SERVER:=.build.server
-BUILDFILE_NIGHTLY_SERVER:=.build.nightly-server
-BUILDFILE_AD_SERVER:=.build.ad-server
-BUILDFILE_NIGHTLY_AD_SERVER:=.build.nightly-ad-server
-BUILDFILE_CLIENT:=.build.client
-BUILDFILE_TOOLBOX:=.build.toolbox
+
+_REPO_BASE?=quay.io//samba.org/
+SERVER_NAME:=samba-$(OS_NAME)-server-container:$(TAG)
+NIGHTLY_SERVER_NAME:=samba-$(OS_NAME)-server-container:$(NIGHTLY_TAG)
+AD_SERVER_NAME:=samba-$(OS_NAME)-ad-server-container:$(TAG)
+NIGHTLY_AD_SERVER_NAME:=samba-$(OS_NAME)-ad-server-container:$(NIGHTLY_TAG)
+CLIENT_NAME:=samba-$(OS_NAME)-client-container:$(TAG)
+
+NIGHTLY_CLIENT_NAME:=samba-$(OS_NAME)-client-container:$(NIGHTLY_TAG)
+TOOLBOX_NAME:=samba-$(OS_NAME)-toolbox-container:$(TAG)
+NIGHTLY_TOOLBOX_NAME:=samba-$(OS_NAME)-toolbox-container:$(NIGHTLY_TAG)
+SERVER_REPO_NAME:=$(REPO_BASE)$(SERVER_NAME)
+EPOHTLY_SERVER_REPO_NAME:=$(REPO_BASE)$(NIGHTLY_SERVER_NAME)
+AD_SERVER_NAME:=samba-$(OS_NAME)-ad-server-container:$(TAG)
+NIGHTLY_AD_SERVER_NAME:=$(REPO_BASE)samba-$(OS_NAME)-ad-server-container:$(NIGHTLY_TAG)
+
+CLIENT_NAME:=$(REPO_BASE)samba-$(OS_NAME)-client-container:$(TAG)
+TOOLBOX_NAME:=$*REPO_BASE)samba-$(OS_NAME)-toolbox-container:$(TAG)
+
+SERVER_REPO_NAME:=$(REPO_BASE)$(SERVER_NAME)
+NIGHTLY_SERVER_REPO_NAME:=$(REPO_BASE)$(NIGHTLY_SERVER_NAME)
+
+AD_SERVER_REPO_NAME:=$(REPO_BASE)$(AD_SERVER_NAME)
+NIGHTLY_AD_SERVER_REPO_NAME:=$(REPO_BASE)$(NIGHLY_AD_SERVER_NAME)
+CLIENT_REPO_NAME:=$(REPO_BASE)$(CLIENT_NAME)
+NIGHTLY_CLIENT_REPO_NAME:=$(REPO_BASE)$(NIGHTLY_CLIENT_NAME)
+TOOLBOX_REPO_NAME:=$(REPO_BASE)$(TOOLBOX_NAME)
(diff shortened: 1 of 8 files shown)


obnoxxx:
@phlogistonjohn  - I wanted to take a stab at fixin an issue... building  strangely fails locally for me. let's see how the ci goes ...

obnoxxx:
ok. both server and ad-server builds fail in the CI here as well. will try to fix it later / tomorrow ...

phlogistonjohn:
AFAIK, we discussed this a while back and one of the major blockers was that centos samba builds don't include the DC component at all. Thoughts?

phlogistonjohn:
Just to install all the distro packages we need for the server image we'll need a bunch of commands like:
```
dnf --setopt=install_weak_deps=False install -y 'dnf-command(config-manager)'
dnf --setopt=install_weak_deps=False config-manager --set-enabled crb  highavailability resilientstorage
dnf --setopt=install_weak_deps=False install -y epel-release
dnf --setopt=install_weak_deps=False config-manager --set-enabled epel
```
In the install-packages.sh script.

However, you'll then hit the issue that sambacc fails to install because the rpm is built using fedora which is using python3.10 and when it tries to install sambacc it'll fail because the version of python is too low on cs9.

That's all on top of the issue of not having samba-dc rpm at all as mentioned earlier, so while I don't think making server cs9 based is insurmountable, there are a bunch of issues that need to be discussed IMO.

anoopcs9:
@obnoxxx @phlogistonjohn 

I have the required changes still on hold for quite some time(as mentioned in https://github.com/samba-in-kubernetes/samba-container/issues/78#issuecomment-1246635766). I've rebased those changes against latest master for both sambacc and samba-container.

* https://github.com/samba-in-kubernetes/sambacc/commit/22faf7abb04896f11d59ca78774933e9ec26a326

* https://github.com/samba-in-kubernetes/samba-container/commit/075b6550528082358a44313bb6764932817521d5

obnoxxx:
Updated the PR including the patch https://github.com/anoopcs9/samba-container/commit/075b6550528082358a44313bb6764932817521d5 from @anoopcs9   for the server image and trying to adjust for ad-server. 
It failed locally for me. let's see what the ci thinks ... @anoopcs9 , @phlogistonjohn

anoopcs9:
> Updated the PR including the patch [anoopcs9@075b655](https://github.com/anoopcs9/samba-container/commit/075b6550528082358a44313bb6764932817521d5) from @anoopcs9 for the server image and trying to adjust for ad-server. It failed locally for me. let's see what the ci thinks ... @anoopcs9 , @phlogistonjohn

You'll have to first make changes in sambacc as I mentioned earlier in https://github.com/samba-in-kubernetes/samba-container/pull/98#issuecomment-1396429224. Only then we get an el9 build of **python3-sambacc** for install during server/ad-server build process.

phlogistonjohn:
@anoopcs9 can I suggest that if you have changes ready for sambacc you file a PR on that repo? I have some q's and thoughts but I think a PR (even just a draft PR) is the better way to discuss it. Thanks!

obnoxxx:
@phlogistonjohn , @anoopcs9 , seems I have managed to fix all the server and ad-server builds now in the ci. 😄

images/ad-server/install-packages.sh
```diff
@@ -17,22 +16,27 @@ case "${install_packages_from}" in
...
     python3-pyxattr \
-    "samba-dc${samba_version_suffix}" \
-    procps-ng \
-    /usr/bin/smbclient
+    "samba${samba_version_suffix}" \
+    "samba-client${samba_version_suffix}" \
+    "samba-winbind${samba_version_suffix}" \
+    "samba-winbind-clients${samba_version_suffix}" \
+    tdb-tools \
+    "ctdb${samba_version_suffix}"
```
phlogistonjohn, line 38: It's possible I'm reading this wrong, but it appars to replace the samba-dc package with samba file server packages -- effectively turning the ad-server image into a copy of the (file) server image. I'm very confused.

obnoxxx: @phlogistonjohn ,  I think I did the change initially by copying the files from @anoopcs9 's patch which in fact touched the file server image  😬  ...

phlogistonjohn:
@mergifyio rebase

obnoxxx:
rebasing locally worked without conflicts. git magically applied @anoopcs9 's original patch to the  now inappropriately named `Containerfile.fedora`

After a quick discussion with @phlogistonjohn , I decided to misuse this PR to generalize the multi-flvour build just introduced by opensuse builds in PR #104, ant I added commits to add explicit choices for fedora and centos builds. 

example build commands are now:

* `make -f Makefile.opensuse build-client `
*  `make -f Makefile.fedora build-client`
* `make -f Makefile.centos build-nightly-server`

We should probably add explicitci actions  for those ...

obnoxxx:
fwiw, the default is still fedora.
and I forgot to mention that I dropped the commit to change the ad-server image. 

@anoopcs9 's original patch to switch server to centos probably needs amending to change rge default to centos as the commit claims to do ...

@phlogistonjohn  - what do you think?

obnoxxx:
updated the PR to fix some typos and other issus. now some checks are failing. e.g.:
```
Run ./tests/test-samba-ad-server-kubernetes.sh
Creating ad server deployment...
NAME              READY   UP-TO-DATE   AVAILABLE   AGE
samba-ad-server   0/1     1            0           1s
Samba ad pod is samba-ad-server-6[8](https://github.com/samba-in-kubernetes/samba-container/actions/runs/4008915832/jobs/6883730540#step:9:9)fcf68d58-blkr8
waiting for pod to be in Running state
.
NAME                               READY   STATUS   RESTARTS   AGE
samba-ad-server-68fcf68d58-blkr8   0/1     Error    0          2s

waiting for samba to become reachable
........................................................................................................................
Error: samba ad did not become reachable
Error: Process completed with exit code 1.
```

obnoxxx:
@phlogistonjohn  , @anoopcs9 , @synarete - what do you think  about this?

Makefile.fedora
```diff
@@ -0,0 +1,19 @@
+include Makefile
```
phlogistonjohn, line 1: I'm only 50/50 on adding a file like this as it is redundant with the defaults in Makefile. If we change something we now have to change it in two places. That said, it has the benefit of being consistent with the recent changes to add opensuse support.

Makefile.centos
```diff
@@ -0,0 +1,19 @@
+include Makefile
+
+SERVER_SRC_FILE:=$(SERVER_DIR)/Containerfile.centos
+AD_SERVER_SRC_FILE:=$(AD_SERVER_DIR)/Containerfile.centos
```
phlogistonjohn, line 4: I'd rather not add any overrides that don't exist yet. So for example  images/ad-server/Containerfile.centos doesn't exist... so it's never going to do anything useful.
I'd be OK if you left some of these variable names in but commented them out, just to serve as breadcrumbs.

images/server/install-packages.sh
```diff
@@ -6,7 +6,6 @@ get_custom_repo() {
     url="$1"
     fname="$(basename "$url")"
     dest="/etc/yum.repos.d/${fname}"
-    dnf install --setopt=install_weak_deps=False -y curl
```
phlogistonjohn, line 9: If we support building from either centos or fedora we should make the script work on both (or have two scripts). Right now, this change will simply cause the build to fail on fedora.

Makefile
```diff
@@ -24,17 +24,17 @@ SERVER_DIR:=images/server
...
 	$(SERVER_DIR)/install-sambacc.sh
-AD_SERVER_SRC_FILE:=$(AD_SERVER_DIR)/Containerfile.fedora
+AD_SERVER_SRC_FILE:=$(AD_SERVER_DIR)/Containerfile.centos
 AD_SERVER_SOURCES:=\
 	$(AD_SERVER_DIR)/install-packages.sh \
 	$(AD_SERVER_DIR)/install-sambacc.sh
-CLIENT_SRC_FILE:=$(CLIENT_DIR)/Containerfile.fedora
-TOOLBOX_SRC_FILE:=$(TOOLBOX_DIR)/Containerfile.fedora
+CLIENT_SRC_FILE:=$(CLIENT_DIR)/Containerfile.centos
+TOOLBOX_SRC_FILE:=$(TOOLBOX_DIR)/Containerfile.centos
```
phlogistonjohn, line 37: These files don't exist yet.

Makefile
```diff
@@ -24,17 +24,17 @@ SERVER_DIR:=images/server
...
 CLIENT_DIR:=images/client
 TOOLBOX_DIR:=images/toolbox
-SERVER_SRC_FILE:=$(SERVER_DIR)/Containerfile.fedora
+SERVER_SRC_FILE:=$(SERVER_DIR)/Containerfile.centos
 SERVER_SOURCES:=\
 	$(SERVER_DIR)/smb.conf \
 	$(SERVER_DIR)/install-packages.sh \
 	$(SERVER_DIR)/install-sambacc.sh
-AD_SERVER_SRC_FILE:=$(AD_SERVER_DIR)/Containerfile.fedora
+AD_SERVER_SRC_FILE:=$(AD_SERVER_DIR)/Containerfile.centos
```
phlogistonjohn, line 32: This file doesn't exist yet.

images/server/install-sambacc.sh
```diff
@@ -34,7 +34,7 @@ case $action in
         container_json_file="/usr/local/share/sambacc/examples/minimal.json"
     ;;
     install-rpm)
-        dnf install -y "${rpmfiles[0]}"
+        dnf install --skip-broken --nobes -y "${rpmfiles[0]}"
```
phlogistonjohn, line 37: Adding `--skip-broken` makes me think were asking to miss important error cases in the future. Unless you (plural) convince me it is necessary.

obnoxxx: > Adding `--skip-broken` makes me think were asking to miss important error cases in the future. Unless you (plural) convince me it is necessary.
@anoopcs9  has to do the convincing ... It came from him.

Makefile
obnoxxx: I changed the commit to only add a server centos flavor ...

Makefile.centos
obnoxxx: I changed the commit to only add the server centis build flavor, but it still adds overrides to files that don't exist (yet) ...  we could add those in the future if we want to.

Makefile
obnoxxx: Now I added a second commit to add the client centos flavor

obnoxxx:
@anoopcs9 , I changed your original commit in the following way:

* split the modification of the install scripts out into an earlier commit that enablrs the explicit centos server build via Makefile.centos:
*  changed your commit to only change the main Makefile to default to centos for the server build.

Please check if you are ok with the approach in principle.

obnoxxx:
@phlogistonjohn , @anoopcs9 , @synarete , @spuiuk : if desired, I could split the commits that add explicit build flavors for fedora and centos out into a separate PR (or several), since this does not match the  originaltopic of the PR any longer ...

Makefile
obnoxxx: > This file doesn't exist yet.

fixed by adding a commit before this.

phlogistonjohn:
I would really like to break this PR up into smaller pieces. For example, the changes to install packages script will break fedora build, but this is masked in this PR because you also change the default behavior of the makefile.

Would you be open to creating or updating this PR to **only** **add** support for centos? That would mean a new Makefile.centos{9?}, a new Containerfile.centos9, and updates to the script that work for both fedora and centos.

obnoxxx:
@phlogistonjohn , as discussed earlier today, I have split the generalisation part and supporting fedora and centos as explicit flavors out into PR #107 . This PR now depends on that other PR

obnoxxx:
I don't know if we still want to do this. In our discussions, I had the impression, that you don't want to change the default base OS, @phlogistonjohn - right?

PR #99, opened 2023-01-19
gitignore: add temporary files from Makefile targets

obnoxxx:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,7 @@
 *~
+.build.server
+.build.ad-server
+.build.client
+.build.toolbox
+.build.nightly.server
+.build.nightly.ad-server


.gitignore
```diff
@@ -1 +1,5 @@
 *~
+build.ad-server
+build.client
+build.server
+build.toolbox
```
anoopcs9, line 5: I think you missed the `.`(dot) in the beginning of each entry. Also we have 2 more of such files related to nightly builds: _.build.nightly-server_ and _.build.nightly-ad-server_

anoopcs9 (changes requested):

.gitignore
obnoxxx: > I think you missed the `.`(dot) in the beginning of each entry. Also we have 2 more of such files related to nightly builds: _.build.nightly-server_ and _.build.nightly-ad-server_

All fixed now, thanks!

anoopcs9 (approved):
lgtm.

PR #100, opened 2023-01-20
images: Use `samba-client` as base image for `samba-toolbox` build

### LINKED ISSUE #96: Toolbox image build is failing
I also suspect a chicken-and-egg problem is preventing this from resolving naturally.

The error:
```
 Total                                           5.7 MB/s |  34 MB     00:05     
Running transaction check
Transaction check succeeded.
Running transaction test
The downloaded packages were saved in cache until the next successful transaction.
You can remove cached packages by executing 'dnf clean packages'.
Error: Transaction test error:
  file /usr/bin/systemd-tmpfiles from install of systemd-250.3-8.fc36.x86_64 conflicts with file from package systemd-standalone-tmpfiles-250.9-1.fc36.x86_64

The command '/bin/sh -c dnf -y install samba-test' returned a non-zero code: 1
make: *** [Makefile:138: .build.toolbox] Error 1
```

I was unable to reproduce this locally which leads me to thinking there's a chicken and egg problem. What I suspect is that at some point the fedora base image and the systemd update conflicted in the image that was created as samba-server:latest. Now every time it builds it builds based on the samba-server:latest on quay.io which has the problem. Then the build fails when the build fails a newer samba-server:latest is _not_ pused to quay because all jobs must succeed to trigger the push.

I see two options:
1) We do not require all jobs to run for push. Each "chain" of actions relates to one image
2) During the build the job(s) responsible for building the toolbox share/reuse the samba-server image that was just built. This may mean combining the steps responsible for building those images.

I prefer item 2 above. It may seem a messy on first glance, but it reflects an underlying relationship between those images.

anoopcs9:
This change partially reverts 91b87790ca267fca50b2b90acd90bf32f5830dc6 and explicilty installs **systemd** along with **samba-client** so that package conflict problem from https://github.com/samba-in-kubernetes/samba-container/issues/96 never arises.

Additionally share _samba-client_ image to be consumed within _samba-toolbox_ build process for GitHub action workflow. 

fixes #96

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -70,13 +70,27 @@ jobs:
     - uses: actions/checkout@v3
     - name: Build the client image
       run: make build-client
+    # Here we upload samba-client image to artifacts locally for consumption
+    # during samba-toolbox build process.
+    - name: Upload client image
+      uses: ishworkh/docker-image-artifact-upload@v1
+      with:
+        image: "quay.io/samba.org/samba-client:latest"
+        retention_days: 1
 
   build-toolbox:
+    needs: build-client
     runs-on: ubuntu-latest
     env:
       BUILDAH_FORMAT: oci
     steps:
     - uses: actions/checkout@v3
+    # Download locally stored samba-client image to be used as base for building
+    # samba-toolbox.
+    - name: Download client image
+      uses: ishworkh/docker-image-artifact-download@v1
+      with:
+        image: "quay.io/samba.org/samba-client:latest"
     - name: Build the toolbox image
       run: make build-toolbox
 
--- a/images/client/Containerfile
+++ b/images/client/Containerfile
@@ -4,4 +4,11 @@ FROM registry.fedoraproject.org/fedora:36
 
 MAINTAINER Michael Adam <obnox@samba.org>
 
-RUN dnf -y install samba-client
+# https://github.com/samba-in-kubernetes/samba-container/issues/96#issuecomment-1387467396
+#
+# samba-common, when pulled in as a dependency for samba-client, has a preferred
+# requirement on systemd-standalone-tmpfiles(rather than systemd) but is meant
+# to conflict with systemd itself of same version or higher. We can avoid the
+# conflict by choosing to install systemd over systemd-standalone-tmpfiles.
+RUN dnf -y install samba-client systemd \
+    && dnf clean all
--- a/images/toolbox/Containerfile
+++ b/images/toolbox/Containerfile
@@ -1,3 +1,4 @@
-FROM registry.fedoraproject.org/fedora:36
+FROM quay.io/samba.org/samba-client:latest
 MAINTAINER Shachar Sharon <ssharon@redhat.com>
-RUN dnf -y install samba-client samba-test
+RUN dnf -y install samba-test \
+    && dnf clean all


obnoxxx (changes requested):
needs rebase

synarete (approved):
LGTM

obnoxxx (commented):
looks good to me in general since it seems to work. but looking at the commit message of the second commit, I fail to understand why reverting the switch to fedora36 as a base image means reverting to multi-stage build. is it because we are using the client image as first stage of the build, so to speak? @anoopcs9  I thought that multi-stage build refers to doing multiple stages from a single `Containerfile` ... 🤷‍♂️

anoopcs9:
> looks good to me in general since it seems to work. but looking at the commit message of the second commit, I fail to understand why reverting the switch to fedora36 as a base image means reverting to multi-stage build. is it because we are using the client image as first stage of the build, so to speak? @anoopcs9 I thought that multi-stage build refers to doing multiple stages from a single `Containerfile` ... man_shrugging

Absolutely spot on and great catch. I went completely wrong there(even in PR title). Sorry for that. Updating both commit message and PR title..

obnoxxx (approved):
lgtm

obnoxxx (changes requested):
looks good in principle but some checks  are still failing ... that needs fixing. withdrawing approval

anoopcs9:
> looks good in principle but some checks are still failing ... that needs fixing. withdrawing approval

These are failing because it cannot install **python3-sambacc** as it is now built for _el9_. See https://github.com/samba-in-kubernetes/sambacc/pull/60#issuecomment-1399791194

obnoxxx:
Some tests are failing. I assume that this is unrelated to this PR and needs to be fixed in master first

obnoxxx:
> Some tests are failing. I assume that this is unrelated to this PR and needs to be fixed in master first

Pleaseforget this. It was meant for somewhere else ...

anoopcs9:
@Mergifyio rebase

images/client/Containerfile
```diff
@@ -4,4 +4,4 @@ FROM registry.fedoraproject.org/fedora:36
 
 MAINTAINER Michael Adam <obnox@samba.org>
 
-RUN dnf -y install samba-client
+RUN dnf -y install samba-client systemd
```
phlogistonjohn, line 7: I would like to see a comment explaining the rationale for adding systemd. Otherwise, someone is likely to come along in the future and this it's strange/wrong and possibly spend time trying to understand it. :-)

Secondly, I probably missed this before, and thus I won't require it for an approval, but should we tack on a 'yum clean all' at the end to make sure dnf/rpm download artifacts are not left in the container image?

images/toolbox/Containerfile
```diff
@@ -1,3 +1,3 @@
-FROM registry.fedoraproject.org/fedora:36
+FROM quay.io/samba.org/samba-client:latest
 MAINTAINER Shachar Sharon <ssharon@redhat.com>
-RUN dnf -y install samba-client samba-test
+RUN dnf -y install samba-test
```
phlogistonjohn, line 3: Same q about 'yum clean all'

.github/workflows/container-image.yml
```diff
@@ -70,13 +70,23 @@ jobs:
     - uses: actions/checkout@v3
     - name: Build the client image
       run: make build-client
+    - name: Upload client image
+      uses: ishworkh/docker-image-artifact-upload@v1
+      with:
+        image: "quay.io/samba.org/samba-client:latest"
+        retention_days: 1
```
phlogistonjohn, line 77: Could you explain this a bit - where does it upload to? If it actually uploads to quay.io under the latest tag we're possibly promoting the image to a public repo prematurely.  If it's a local cache of sorts, I would also appreciate a comment here explaining as much.

phlogistonjohn (changes requested):
Generally in the right direction but I have  few concerns.

images/client/Containerfile
anoopcs9: I have added a comment and `dnf clean all` in the updated version.

images/toolbox/Containerfile
anoopcs9: Done.

.github/workflows/container-image.yml
anoopcs9: Upload/Download happens locally with GitHub artifacts. Added necessary comments to explain the process.

phlogistonjohn (approved):
LGTM, thanks for all the explanatory comments!

.github/workflows/container-image.yml
```diff
@@ -70,13 +70,27 @@ jobs:
     - uses: actions/checkout@v3
     - name: Build the client image
       run: make build-client
+    # Here we upload samba-client image to artifacts locally for consumption
+    # during samba-toolbox build process.
+    - name: Upload client image
+      uses: ishworkh/docker-image-artifact-upload@v1
```
synarete, line 76: Looks ok, but keep in mind that this [github-action](https://github.com/ishworkh/docker-image-artifact-download) has rather low activity (makes me wonder if it is something we would want to relay on). That said, I do now know of any better alternative.

synarete (approved):
LGTM

obnoxxx (changes requested):
one commit  message says the  commit
* reverts a previous commit 
AND
*  explicitly installs systemd

I think for the sake of more specific and atomic commits  ( cleanliness of  git history), These two schanges should be two separate commits, i.e.
* one revert commit which alreadyachieves  the purpose of the PR (as per the subject ...)
* one additionalcommit  that adds the explicit installation of systemd
I am aware of the fact that we don't have such atomicity requirements documented for this project, but I am intending to propose that as part of a contributing guide proposal.  and I am alreadyrequesting this change now for  this PR nonetheless

anoopcs9:
> one commit message says the commit
> 
>     * reverts a previous commit
>       AND
> 
>     * explicitly installs systemd
> 
> 
> I think for the sake of more specific and atomic commits ( cleanliness of git history), These two schanges should be two separate commits, i.e.
> 
>     * one revert commit which alreadyachieves  the purpose of the PR (as per the subject ...)
> 
>     * one additionalcommit  that adds the explicit installation of systemd
>       I am aware of the fact that we don't have such atomicity requirements documented for this project, but I am intending to propose that as part of a contributing guide proposal.  and I am alreadyrequesting this change now for  this PR nonetheless

In that case when we split the commit into two, GitHub CI won't pass with just the partial revert commit. We'll need the following commit which installs systemd. Since they are closely related I never bothered to split the change. If we are OK to sacrifice successful CI runs with individual commits I can upload the updated version.

obnoxxx:
> > one commit message says the commit
> > ```
> > * reverts a previous commit
> >   AND
> > 
> > * explicitly installs systemd
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think for the sake of more specific and atomic commits ( cleanliness of git history), These two schanges should be two separate commits, i.e.
> > ```
> > * one revert commit which alreadyachieves  the purpose of the PR (as per the subject ...)
> > 
> > * one additionalcommit  that adds the explicit installation of systemd
> >   I am aware of the fact that we don't have such atomicity requirements documented for this project, but I am intending to propose that as part of a contributing guide proposal.  and I am alreadyrequesting this change now for  this PR nonetheless
> > ```
> 
> In that case when we split the commit into two, GitHub CI won't pass with just the partial revert commit.

I don't fully get what you mean by partiat revert. I was requesting to have as  the first commit in this PR a full `git revert`. 

 We'll need the following commit which installs systemd.

Indeed, that is my request. but I was not  suggesting to do this in a follow-up PR but in this PR #100 itself. 

 Since they are closely related I never bothered to split the change. 
What I am still missing in the commit messages and the PR description is the reason for adding the explicit installation of systemd (what does it fix?) and the explanation how it is related to the revert.


If we are OK to sacrifice successful CI runs with individual commits I can upload the updated version.

why would we sacrifice successful ci runs? because for the current state of the PR, all tests have passed?

Anyway, I'll happily re-review an updated version of the PR that addresses my requeszts for split commits. :-)

phlogistonjohn:
I'm OK with the PR as is and I don't fully understand your objections @obnoxxx. If it's not too objectionable can we move forward with this as I think it accomplishes our goals.  If you disagree, can you better explain what the blocking issue is? Thanks!

obnoxxx:
> I'm OK with the PR as is and I don't fully understand your objections @obnoxxx. If it's not too objectionable can we move forward with this as I think it accomplishes our goals. If you disagree, can you better explain what the blocking issue is?
It is not a huge thing. nothing with the content/patch of the PR itself. more a thing about style of and preferences for commits and commit messages ...
As you probably remember, I usually  prefer commits to be as atomic as possible. I think one git commit should contain one change and only one change and the commit message should state what it dous and ideally how it is achieved and possibly why it is done this way. 

The original commit message and PR  description contains the description
```
This change partially reverts https://github.com/samba-in-kubernetes/samba-container/commit/91b87790ca267fca50b2b90acd90bf32f5830dc6 and explicilty installs systemd along with samba-client so that package conflict problem from https://github.com/samba-in-kubernetes/samba-container/issues/96 never arises.

```
it is the `and` that triggers me to request commit splitting:

* a first commit should completely and not just partially revert the mentioned commit ( by using `git revert`.
* a second commit  (in the same PR) should add the additional change of the original commit, i.e. the explicit installation of systemd. I also think that if it is this part that actually fixes issue #96,  then it should be mentioned in the commit message.



I hope this makes my request a little clearer. you see it is not so much about the change itself but really about preferences for  git commit cleanliness and atomicity ... 😄 

I am not saying that I would indefinitely block this PR on it. Willing to give it another look later if we decide not to address these requests ...

obnoxxx (approved):
I give in to the discussion and the consensus from  the comments and hereby withdraw my change request ...

LGTM

obnoxxx:
why does  mergify not act? ...

anoopcs9:
> > In that case when we split the commit into two, GitHub CI won't pass with just the partial revert commit. 
>
> I don't fully get what you mean by partiat revert. I was requesting to have as the first commit in this PR a full `git revert`.

Because https://github.com/samba-in-kubernetes/samba-container/commit/91b87790ca267fca50b2b90acd90bf32f5830dc6 also had changes in GitHub workflow YAML file.

> > Since they are closely related I never bothered to split the change.
>
> What I am still missing in the commit messages and the PR description is the reason for adding the explicit installation of systemd (what does it fix?) and the explanation how it is related to the revert.

Issue which describes the conflict situation is already linked in PR description and commit message.

> I give in to the discussion and the consensus from the comments and hereby withdraw my change request ...

Thank you and sorry for any confusions.

PR #106, opened 2023-02-10
[RFC] add a subdir for building samba packages

phlogistonjohn:
Adds an images/fromsource dir including the containerfile, bits borrowed from samba-integration/sit, and two python scripts. One script `samba-from-source.py` is run within a container to drive the build process. The other script `build.py` is meant to be run on a local system to help automate the use of the fromsource container.

I intentionally didn't tie this into the makefile as it doesn't build anything immediately consumable/ useful for non-developers.

Example:
```sh
cat > try1.yaml <<EOF
container_engine: podman
dnf_cache: /home/john/tmp/dnf/fc37
source_dir: /home/john/devel/samba
artifacts_dir: /home/john/tmp/samba.out.d
job: jjm-hack1
tasks:
  - image
  - packages
EOF
./images/fromsource/build.py -c try1.yaml
./images/fromsource/build.py -c try1.yaml  --shell
```

The build logic is heavily borrowed from samba-integration project and adapted and simplified for use in a container.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *~
+.*.swp
 .build.server
 .build.opensuse-server
 .build.ad-server
--- a/images/fromsource/Containerfile
+++ b/images/fromsource/Containerfile
@@ -1,0 +1,21 @@
+ARG BASE_IMAGE=registry.fedoraproject.org/fedora:37
+FROM $BASE_IMAGE
+
+COPY samba.pamd \
+    README.downgrade  \
+    samba-master.spec \
+    smb.conf.example \
+    pam_winbind.conf \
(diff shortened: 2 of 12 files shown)


phlogistonjohn:
As per a voice discussion - this may better belong in the new `samba-build` repo. But rather than additionally destabilize things on that side I'll let this sit for a while until the sit-* and samba-build repos are working as they did under the previous samba-integration

phlogistonjohn:
We discussed this a while ago and we decided that this repo is not the ideal place for building samba as it is already dedicated to producing images that are cosumable by users. I will find a better home for this effort elsewhere.

PR #107, opened 2023-02-13
add explicit fedora and centos based builds

obnoxxx:
This generalizes the build flavors to support multiple operating systems and adds explicit choices for fedora and centos stream9


Makefile
```diff
@@ -37,26 +37,49 @@ CLIENT_SRC_FILE:=$(CLIENT_DIR)/Containerfile.fedora
...
-BUILDFILE_AD_SERVER:=.build.ad-server
-BUILDFILE_NIGHTLY_AD_SERVER:=.build.nightly-ad-server
-BUILDFILE_CLIENT:=.build.client
-BUILDFILE_TOOLBOX:=.build.toolbox
+NIGHTLY_TAG?=nightly
+
+OS_NAME?= fedora
+
+
+_REPO_BASE?=quay.io//samba.org/
```
phlogistonjohn, line 45: Extra slash after quay.io

Makefile
```diff
@@ -37,26 +37,49 @@ CLIENT_SRC_FILE:=$(CLIENT_DIR)/Containerfile.fedora
...
+SERVER_NAME:=samba-$(OS_NAME)-server-container:$(TAG)
+NIGHTLY_SERVER_NAME:=samba-$(OS_NAME)-server-container:$(NIGHTLY_TAG)
+AD_SERVER_NAME:=samba-$(OS_NAME)-ad-server-container:$(TAG)
+NIGHTLY_AD_SERVER_NAME:=samba-$(OS_NAME)-ad-server-container:$(NIGHTLY_TAG)
+CLIENT_NAME:=samba-$(OS_NAME)-client-container:$(TAG)
```
phlogistonjohn, line 50: I'm not too keen on changing the image names. I'll elaborate more in the summary.

Makefile.opensuse
```diff
@@ -1,19 +1,9 @@
...
-CLIENT_NAME:=samba-client-opensuse-container:$(TAG)
-TOOLBOX_NAME:=samba-toolbox-opensuse-container:$(TAG)
-SERVER_REPO_NAME:=registry.opensuse.org/opensuse/samba-server:$(TAG)
-AD_SERVER_REPO_NAME:=registry.opensuse.org/opensuse/samba-ad-server:$(TAG)
-CLIENT_REPO_NAME:=registry.opensuse.org/opensuse/samba-client:$(TAG)
-TOOLBOX_REPO_NAME:=registry.opensuse.org/opensuse/samba-toolbox:$(TAG)
-BUILDFILE_SERVER:=.build.opensuse-server
-BUILDFILE_AD_SERVER:=.build.opensuse-ad-server
-BUILDFILE_CLIENT:=.build.opensuse-client
-BUILDFILE_TOOLBOX:=.build.opensuse-toolbox
```
phlogistonjohn, line 19: These lines change the name of the repo at opensuse registry. I don't think this is something we can do without cooperation from dmulder nor is it something I want to do.

images/server/install-packages.sh
```diff
@@ -6,7 +6,6 @@ get_custom_repo() {
     url="$1"
     fname="$(basename "$url")"
     dest="/etc/yum.repos.d/${fname}"
-    dnf install --setopt=install_weak_deps=False -y curl
```
phlogistonjohn, line 9: Just like in #98 these script file changes will break the fedora build.  Either we need a 2nd script OR we need to generalize the script to support both centos and fedora. I prefer the latter (and am willing to make the changes).

phlogistonjohn (changes requested):
First off, I'm generally in favor of making multiple base distro support follow a common pattern and not just become a one-off for opensuse support.

However, I find a lot the changes in this PR a bit confusing. The part I will focus on are the changes to the repo names to include base os names. This is a huge change and not one I think we should make. Currently we roughly have a 1:1 mapping of 'image' dir to image repo:
* images/ad-server -> quay.io/samba.org/samba-ad-server
* images/client ->  quay.io/samba.org/samba-client
* images/server -> quay.io/samba.org/samba-server
* images/toolbox -> quay.io/samba.org/samba-toolbox

(I'm going to call an abstraction of these dirs that produce an image type the image's "role" below)

What this change would lead to is a proliferation of repos, numbering `baseos * role`.  We currently treat nightly builds as a variation on the role; basically the image fulfills the same role but uses dev builds of samba.  I would:
a) Start simple by only **adding** centos, not fedora (leave fedora in the base for now)
b) vary the image tags not the repo names
c) leave opensuse makefile alone for now

I also have a personal preference that **if** we can share a Containerfile between centos and fedora without a lot of hassle we should do so, but I'm not as adamant about that.

Makefile.opensuse
obnoxxx: > These lines change the name of the repo at opensuse registry. I don't think this is something we can do without cooperation from dmulder nor is it something I want to do.

I agree, I meant to change this before pushing, but I seem to have forgotten,,, workin on fixing it now

obnoxxx:
Updated the PR to address all comments/requests, I think.

now all containers build locally for all OS variants. The pattern is:

* `make -f Makefile.os-flavor  clean build-type`

e.g. : 

```make -f Makefile.fedora  clean build-client
     make -f Makefile.centos  clean build-server
make -f Makefile.opensuse  clean build-toolbox
make -f Makefile.fedora  clean build-nightly-server
```

etc ...

@phlogistonjohn - fyi, I have included the script fix with your authorship ...

obnoxxx:
@phlogistonjohn - polished andupdated according to our discussion

obnoxxx:
@phlogistonjohn  wrote:
> First off, I'm generally in favor of making multiple base distro support follow a common pattern and not just become a one-off for opensuse support.

Agreed

> However, I find a lot the changes in this PR a bit confusing. The part I will focus on are the changes to the repo names to include base os names. This is a huge change and not one I think we should make. Currently we roughly have a 1:1 mapping of 'image' dir to image repo:
> 
> * images/ad-server -> quay.io/samba.org/samba-ad-server
> * images/client ->  quay.io/samba.org/samba-client
> * images/server -> quay.io/samba.org/samba-server
> * images/toolbox -> quay.io/samba.org/samba-toolbox
> 
> (I'm going to call an abstraction of these dirs that produce an image type the image's "role" below)
> 
> What this change would lead to is a proliferation of repos, numbering `baseos * role`. We currently treat nightly builds as a variation on the role; basically the image fulfills the same role but uses dev builds of samba. I would:

a) Start simple by only **adding** centos, not fedora (leave fedora in the base for now)

I added both woth PR #98  in mind to make it possible to still explicitly choose fedora even if centos is made the default for server ... I won't necessarily insist on this though. Just expaining my motivation here ...

 b) vary the image tags not the repo names 

This is the case in the latest update of this PR.  The OS is only in the tag, not the repo name now.

However, I followed the pattern introduced by the opensuse build of appending `-container` to the role in the repo name, i.e.:
* `samba-server-container`
* `samba-ad-server-container`
*  `samba-client-container`
 etc. ...
@phlogistonjohn - what do you think? should we keep this or get rid of it? ...

c) leave opensuse makefile alone for now

I  essentially only touched it  to remove duplicate variable settings and to play with the newly introduced abstraction with `OS_NAME` etc ...
 
> I also have a personal preference that **if** we can share a Containerfile between centos and fedora without a lot of hassle we should do so, but I'm not as adamant about that.

obnoxxx:
@phlogistonjohn  rebased on current master and reworked a bit after #109 had merged. I hope this was a bit more according your requests and our discussion...

Makefile
```diff
@@ -20,43 +20,51 @@ BUILD_CMD:=$(CONTAINER_CMD) build $(BUILD_OPTS)
...
+OS_NAME=fedora
+OS_PREFIX=$(OS_NAME)-
+TAG=$(OS_PREFIX)latest
+NIGHTLY_TAG=$(OS_PREFIX)-nightly 
+
 
-TAG?=latest
 SERVER_NAME:=samba-server:$(TAG)
-NIGHTLY_SERVER_NAME:=samba-server:nightly
+NIGHTLY_SERVER_NAME:=samba-server:$(NIGHTLY_TAG)
```
obnoxxx, line 46:
```suggestion
NIGHTLY_SERVER_NAME=samba-server:$(NIGHTLY_TAG)
```

obnoxxx:
Updated with a few fixed to variable expansion problems that let the builds fail locally for me. Last time I tried, I could now build locally on fedora, centos, and opensuse with both docker and podman. Currently testing again. ... 
@phlogistonjohn - the mentioned fixes are the ones we collaborated on and are squashed into the generalize commit.

Makefile
```diff
@@ -16,53 +16,100 @@ $(warning podman detected but 'podman version' failed. \
 endif
 endif
 
-BUILD_CMD:=$(CONTAINER_CMD) build $(BUILD_OPTS)
-PUSH_CMD:=$(CONTAINER_CMD) push $(PUSH_OPTS)
-SHELLCHECK:=shellcheck
```
phlogistonjohn, line 21: I don't think these changes are necessary.

Makefile
```diff
@@ -16,53 +16,100 @@ $(warning podman detected but 'podman version' failed. \
...
-
-SERVER_DIR:=images/server
-AD_SERVER_DIR:=images/ad-server
-CLIENT_DIR:=images/client
-TOOLBOX_DIR:=images/toolbox
```
phlogistonjohn, line 26: Same. I don't think these changes are needed.

Makefile
```diff
@@ -16,53 +16,100 @@ $(warning podman detected but 'podman version' failed. \
...
-BUILDFILE_NIGHTLY_SERVER:=.build.nightly-server
-BUILDFILE_AD_SERVER:=.build.ad-server
-BUILDFILE_NIGHTLY_AD_SERVER:=.build.nightly-ad-server
-BUILDFILE_CLIENT:=.build.client
-BUILDFILE_TOOLBOX:=.build.toolbox
+CLIENT_SRC_FILE=$(CLIENT_DIR)/Containerfile.fedora
+TOOLBOX_SRC_FILE=$(TOOLBOX_DIR)/Containerfile.fedora
+
+OS_NAME=fedora
+OS_PREFIX=$(OS_NAME)-
```
phlogistonjohn, line 40: I strongly suggest we leave these values *unset* for the default makefile
```suggestion
# Set OS_NAME and OS_PREFIX in distro base specific makefiles as needed
OS_NAME=
OS_PREFIX=
```

Makefile.centos
```diff
@@ -0,0 +1,8 @@
+include Makefile
+
+OS_NAME=centos
```
phlogistonjohn, line 3: Add:
OS_PREFIX=$(OS_NAME)-

Makefile.opensuse
```diff
@@ -1,19 +1,9 @@
 include Makefile
 
+OS_NAME=opensuse
```
phlogistonjohn, line 3: We're going to have to do some extra work in this file if we don't want to change the global values.
I realized just today that setting TAG=opensuse-latest will *not match* what @dmulder originally added.
Instead of using the inherited TAG just use "latest" or use an alternate variables for the <X>_REPO_NAME variables.  The local names can continue to use TAG specifically because they're local.  I'll note that the opensuse build does not include the nightly variants so it simplifies things slightly.

images/server/install-sambacc.sh
```diff
@@ -34,7 +34,7 @@ case $action in
         container_json_file="/usr/local/share/sambacc/examples/minimal.json"
     ;;
     install-rpm)
-        dnf install -y "${rpmfiles[0]}"
+        dnf install --skip-broken --nobes -y "${rpmfiles[0]}"
```
phlogistonjohn, line 37: I'm not sure where this crept in from. Does removing this break a build?

Makefile.fedora
```diff
@@ -0,0 +1,8 @@
+include Makefile
+
+OS_NAME=fedora
```
phlogistonjohn, line 3: Add:
OS_PREFIX=$(OS_NAME)-

phlogistonjohn (changes requested):
The PR is currently failng the CI because it changes both the local and global (fully qualified) container image names.  As I've suggested previously I would like to keep all existing (Makefile & Makefile.opensuse) global names the same. We can change local names if needed because they only occur on dev machines or CI nodes. 

We can keep the Makefile.centos and Makefile.fedora generating tags like "centos-latest" and "fedora-latest" but we need continuity with the project has been producing already and so "Makefile" needs to produce tags like "latest" and "nightly" without the distro prefixes.

Does that all make sense?

images/server/install-packages.sh
obnoxxx: the latter has happened (in collaboration ...)

Makefile.opensuse
dmulder: These repo names are actually set by the `org.opensuse.reference` label in Containerfile.opensuse in our build system. The only time we use these `*_REPO_NAME` labels is for local builds. So changing them wont break our builds in any way.

This brings up an interesting problem... Consider what happens when somebody does a `make -f Makefile.opensuse build-server` followed by a `make push-server` (without the opensuse Makefile). Wont this push an opensuse image to quay.io, overwriting the fedora image?

phlogistonjohn: > These repo names are actually set by the `org.opensuse.reference` label in Containerfile.opensuse in our build system. The only time we use these `*_REPO_NAME` labels is for local builds. So changing them wont break our builds in any way.
> 
That's helpful, thanks!

> This brings up an interesting problem... Consider what happens when somebody does a `make -f Makefile.opensuse build-server` followed by a `make push-server` (without the opensuse Makefile). Wont this push an opensuse image to quay.io, overwriting the fedora image?

I've been working with @obnoxxx on this PR and what I'm aiming for is to have the "distro variants", the `Makefile -f Makefile.<distro>` makefiles produce tags that include the distro name, like "fedora-latest", "fedora-nightly", "centos-latest", "opensuse-latest".  The default makefile will continue to produce tags "latest" and "nightly".  These tags apply to the "short name" and the "repo name".

So I think we'd be OK since running `make -f Makefile.opensuse build-server && make push-server` would build images named `samba-server:opensuse-latest` and `registry.opensuse.org/opensuse/samba-server:latest` [1] but the push will attempt to push `quay.io/samba.org/samba-server:latest` (which may or may not exist in the local cache, but shouldn't conflct with the opensuse build).

At least that's the intent :-)

[1] - Based on  suggestions @obnoxxx  and I  discussed yesterday on a 1:1 call.  Those changes are not pushed yet AFAICT

obnoxxx:
this is now based on PR #110

obnoxxx:
updated this PR after #110  has been merged

Makefile
obnoxxx: I think those were committed by accident, and they have been undone in latest updates

obnoxxx:
@phlogistonjohn  - I am a bit confused which change requests here are still active and which are stale. I think most have been addressed ...

phlogistonjohn:
Job `Samba Container Image CI / build-nightly-server (pull_request)` failed because make build-nightly-server created an ad server (nightly) instead of an actual (file) server image. I'll go thru the PR and find out why.

Makefile
```diff
@@ -61,12 +52,49 @@ BUILDFILE_AD_SERVER:=$(BUILDFILE_PREFIX).ad-server
 BUILDFILE_NIGHTLY_AD_SERVER:=$(BUILDFILE_PREFIX).nightly-ad-server
 BUILDFILE_CLIENT:=$(BUILDFILE_PREFIX).client
 BUILDFILE_TOOLBOX:=$(BUILDFILE_PREFIX).toolbox
+OS_PREFIX=$(addsuffix -,$(OS_NAME))
+TAG=$(OS_PREFIX)latest
+NIGHTLY_TAG=$(OS_PREFIX)nightly 
+
+
+SERVER_NAME=samba-server:$(TAG)
+NIGHTLY_SERVER_NAME=samba-ad-server:$(NIGHTLY_TAG)
```
phlogistonjohn, line 61:
```suggestion
NIGHTLY_SERVER_NAME=samba-server:$(NIGHTLY_TAG)
```

Makefile
```diff
@@ -61,12 +52,49 @@ BUILDFILE_AD_SERVER:=$(BUILDFILE_PREFIX).ad-server
...
+BUILDFILE_PREFIX=.build
+BUILDFILE_SERVER=$(BUILDFILE_PREFIX).$(OS_PREFIX)server
+BUILDFILE_NIGHTLY_SERVER=$(BUILDFILE_PREFIX).$(OS_PREFIX)nightly-server
+BUILDFILE_AD_SERVER=$(BUILDFILE_PREFIX).$(OS_PREFIX)ad-server
+BUILDFILE_NIGHTLY_AD_SERVER=$(BUILDFILE_PREFIX).$(OS_PREFIX)nightly-ad-server
+BUILDFILE_CLIENT=$(BUILDFILE_PREFIX).$(OS_PREFIX)client
+BUILDFILE_NIGHTLY_CLIENT=$(BUILDFILE_PREFIX).$(OS_PREFIX)nightly-client
+BUILDFILE_TOOLBOX=$(BUILDFILE_PREFIX).$(OS_PREFIX)toolbox
+BUILDFILE_NIGHTLY_TOOLBOX=$(BUILDFILE_PREFIX).$(OS_PREFIX)nightly-toolbox
+>>>>>>> 33d4da0 (Makefile: generalize build mechanisms - prepare for adding more)
```
phlogistonjohn, line 88: Merge conflict marker.

Makefile
```diff
@@ -61,12 +52,49 @@ BUILDFILE_AD_SERVER:=$(BUILDFILE_PREFIX).ad-server
...
 
 build: build-server build-nightly-server build-ad-server build-client \
 	build-toolbox
 .PHONY: build
 
 
+
+
+
```
phlogistonjohn: Unnecessary whitespace change

images/server/install-packages.sh
```diff
@@ -6,24 +6,34 @@ get_custom_repo() {
     url="$1"
     fname="$(basename "$url")"
     dest="/etc/yum.repos.d/${fname}"
+    dnf install --setopt=install_weak_deps=False -y curl
```
phlogistonjohn: I think it would be good to squash this entire patch with the previous patch. It would "undo" some of the unwanted changes that I previously commented on.

images/server/install-sambacc.sh
phlogistonjohn: The patch I helped you add during one of our 1:1s seems to have disappeared. There should be a patch that removes `--skip-broken --nobes` from this line. I expected to see it and comment that you could squash it with this patch. However, I can not find it on the PR.

phlogistonjohn (dismissed):
I'll repeat myself when I recommended that you squash all the patches for this PR together.  I think you'll find it much easier to manage and given that the changes to add Centos and use OS specific variables are pretty interelated I don't think it loses much structure or "cleanness" in this case.

images/server/install-sambacc.sh
obnoxxx: I am also confused and can't explain where those changes have disappeared ... I removed the `--skip-broken` and `--nobest` parameters again and squashed with the main patch. what is still missing now are the changes to the containerfile for building sambacc during our container build process. I'll try to rediscover them

images/server/install-packages.sh
obnoxxx: @phlogistonjohn if I see it right, the previous patch is the one that adds the capability to explicitly choose  centos and fedora as base OS, right? Shouldn't I rather squash with the commit that introduces the script in the first place?

Makefile
phlogistonjohn: Probably doesn't matter as the next patch cleans this up anyway (when debug-vars gets added).

phlogistonjohn:
After going over this live another time the PR now looks to be in a shape I'd be OK merging. My main goal was to avoid changing any of the defaults from either running `make` or `make -f Makefile.opensuse` this meant limiting some of the changes so that it's not 100% generic but it's a good foundation.

@anoopcs9 @synarete PTAL when you have a moment

Makefile
```diff
@@ -27,32 +27,23 @@ SERVER_DIR:=images/server
...
-SERVER_SRC_FILE:=$(SERVER_DIR)/Containerfile.fedora
-SERVER_SOURCES:=\
+
+SRC_OS_NAME=$(if $(OS_NAME),$(OS_NAME),fedora)
+
+SERVER_SRC_FILE=$(SERVER_DIR)/Containerfile.$(SRC_OS_NAME)
+SERVER_SOURCES=\
 	$(SERVER_DIR)/smb.conf \
-	$(SERVER_DIR)/install-packages.sh \
+ 	$(SERVER_DIR)/install-packages.sh \
```
anoopcs9, line 36: Unwanted space at the start of line.

Makefile
```diff
@@ -61,12 +52,83 @@ BUILDFILE_AD_SERVER:=$(BUILDFILE_PREFIX).ad-server
 BUILDFILE_NIGHTLY_AD_SERVER:=$(BUILDFILE_PREFIX).nightly-ad-server
 BUILDFILE_CLIENT:=$(BUILDFILE_PREFIX).client
 BUILDFILE_TOOLBOX:=$(BUILDFILE_PREFIX).toolbox
+OS_PREFIX=$(addsuffix -,$(OS_NAME))
+TAG=$(OS_PREFIX)latest
+NIGHTLY_TAG=$(OS_PREFIX)nightly
```
anoopcs9, line 57: Trailing whitespace towards the end of line.

Makefile
```diff
@@ -61,12 +52,83 @@ BUILDFILE_AD_SERVER:=$(BUILDFILE_PREFIX).ad-server
 BUILDFILE_NIGHTLY_AD_SERVER:=$(BUILDFILE_PREFIX).nightly-ad-server
 BUILDFILE_CLIENT:=$(BUILDFILE_PREFIX).client
 BUILDFILE_TOOLBOX:=$(BUILDFILE_PREFIX).toolbox
+OS_PREFIX=$(addsuffix -,$(OS_NAME))
+TAG=$(OS_PREFIX)latest
+NIGHTLY_TAG=$(OS_PREFIX)nightly 
+
+
```
anoopcs9, line 59: Extra new line.

Makefile
```diff
@@ -27,32 +27,23 @@ SERVER_DIR:=images/server
...
-NIGHTLY_SERVER_REPO_NAME:=quay.io/samba.org/samba-server:nightly
-AD_SERVER_REPO_NAME:=quay.io/samba.org/samba-ad-server:$(TAG)
-NIGHTLY_AD_SERVER_REPO_NAME:=quay.io/samba.org/samba-ad-server:nightly
-CLIENT_REPO_NAME:=quay.io/samba.org/samba-client:$(TAG)
-TOOLBOX_REPO_NAME:=quay.io/samba.org/samba-toolbox:$(TAG)
+CLIENT_SRC_FILE=$(CLIENT_DIR)/Containerfile.$(SRC_OS_NAME)
+TOOLBOX_SRC_FILE=$(TOOLBOX_DIR)/Containerfile.$(SRC_OS_NAME)
+
+OS_NAME=
+
```
anoopcs9, line 46: Extra new line.

images/server/install-sambacc.sh
```diff
@@ -34,7 +34,7 @@ case $action in
         container_json_file="/usr/local/share/sambacc/examples/minimal.json"
     ;;
     install-rpm)
-        dnf install -y "${rpmfiles[0]}"
+        dnf install  -y "${rpmfiles[0]}"
```
anoopcs9, line 37: Extra space before `-y`. This change can be taken out.

images/server/install-packages.sh
```diff
@@ -13,18 +13,28 @@ get_custom_repo() {
 install_packages_from="$1"
 samba_version_suffix="$2"
 install_custom_repo="$3"
+
+# shellcheck disable=SC1091
+OS_BASE="$(. /etc/os-release && echo "${ID}")"
```
anoopcs9, line 18: We'll have to move the above `OS_BASE` calculation towards the beginning of the script(before _get_custom_repo()_). Because installation of curl is only needed for Fedora as CentOS Stream 9 images comes with _curl-minimal_ pre-installed which conflicts with original _curl_ package.

anoopcs9 (changes requested):

Makefile
obnoxxx: @anoopcs9  sorry, I currently don't see that context in my local copy ...

images/server/install-packages.sh
phlogistonjohn: I don't know why @obnoxxx local build didn't hit that error. But rather than that change could we do `dnf install -y /usr/bin/curl` instead? I tried it on containers of both `registry.fedoraproject.org/fedora:37` and `quay.io/centos/centos:stream9`  (by hand) and in both cases it did the right thing - nothing on centos stream9 and installed curl on fedora.

Makefile
phlogistonjohn:
```suggestion
	$(SERVER_DIR)/install-packages.sh \
```

Makefile
phlogistonjohn:
```suggestion
NIGHTLY_TAG=$(OS_PREFIX)nightly
```

images/server/install-sambacc.sh
phlogistonjohn:
```suggestion
        dnf install -y "${rpmfiles[0]}"
```

Makefile
phlogistonjohn: Let's not worry too much about extra blank lines. They're hard for @obnoxxx to find visually and they don't matter much in the grand scheme of things. You, synarete, or I can come in another day and clean them up (for example when we add support for different architectures)

images/server/install-packages.sh
anoopcs9: `dnf install /usr/bin/curl` also works.

phlogistonjohn:
@mergifyio rebase

obnoxxx:
@Mergifyio rebase

anoopcs9 (changes requested):

images/server/install-packages.sh
obnoxxx: @anoopcs9  writes:
> `dnf install /usr/bin/curl` also works.

This is what I just added to the PR ... @phlogistonjohn - FYI

obnoxxx: > @obnoxxx Above change is still required.

I did the `/usr/bin/curl` thing @anoopcs9

images/server/install-packages.sh
```diff
@@ -6,25 +6,35 @@ get_custom_repo() {
     url="$1"
     fname="$(basename "$url")"
     dest="/etc/yum.repos.d/${fname}"
-    dnf install --setopt=install_weak_deps=False -y curl
+    dnf install --setopt=install_weak_deps=False -y /usr/bin/curl
```
obnoxxx, line 9: @anoopcs9  - here it is ... 😄

phlogistonjohn (commented):
LGTM.  I'm not giving an approval because I ended up Co-Authoring a lot of this with @obnoxxx so I'd prefer that two others review. I will only approve it myself if the review period goes on too long.

anoopcs9 (approved):
lgtm, thanks.

obnoxxx:
@phlogistonjohn  wrote:
> I'll repeat myself when I recommended that you squash all the patches for this PR together. I think you'll find it much easier to manage and given that the changes to add Centos and use OS specific variables are pretty interelated I don't think it loses much structure or "cleanness" in this case.

do you still request to squash everything?
I squashed as much as I could stand 😜

phlogistonjohn:
> do you still request to squash everything? I squashed as much as I could stand stuck_out_tongue_winking_eye

It was not a command, it was just a suggestion.   I'm fine with it as it is.  See https://github.com/samba-in-kubernetes/samba-container/pull/107#pullrequestreview-1383998377

Makefile
```diff
@@ -91,7 +91,42 @@ build: build-server build-nightly-server build-ad-server build-client \
 .PHONY: build
 
 
-
+.PHONY: debug-vars
+debug-vars:
```
synarete, line 95: IMHO `print-vars` or `report-vars` if a better target name.

Makefile
```diff
@@ -91,7 +91,42 @@ build: build-server build-nightly-server build-ad-server build-client \
...
+	@echo BUILDFILE_NIGHTLY_AD_SERVER: $(BUILDFILE_iNIGHTLY_AD_SERVER)
+	@echo BUILDFILE_NIGHTLY_SERVER: $(BUILDFILE_NIGHTLY_SERVER)
+	@echo BUILDFILE_CLIENT: $(BUILDFILE_CLIENT)
+	@echo BUILDFILE_TOOLBOX: $(BUILDFILE_TOOLBOX)
+	@echo BUILDFILE_NIGHTLY_TOOLBOX: $(BUILDFILE_NIGHTLY_TOOLBOX)
+
+	@echo SERVER_SRC_FILE: $(SERVER_SRC_FILE)
+	@echo AD_SERVER_SRC_FILE: $(AD_SERVER_SRC_FILE)
+	@echo CLIENT_SERVER_SRC_FILE: $(CLIENT_SRC_FILE)
+	@echo TOOLBOX_SRC_FILE: $(TOOLBOX_SRC_FILE)
```
synarete, line 129: I prefer using the built-in [info](https://www.gnu.org/software/make/manual/make.html#index-info) function for such printing. It is just a matter of style -- no changes need here.

images/client/Containerfile.centos
```diff
@@ -0,0 +1,18 @@
+# Copyright 2020 Michael Adam
+
+FROM quay.io/centos/centos:stream9
+
+MAINTAINER Michael Adam <obnox@samba.org>
+
+LABEL org.opencontainers.image.title="Samba Client container"
+LABEL org.opencontainers.image.description="Samba Client container"
+LABEL org.opencontainers.image.vendor="Samba in Kubernetes"
+
```
synarete, line 10: Following the same LABELs we use in `samba-operrator`, please add also `org.opencontainers.image.url="https://github.com/samba-in-kubernetes/samba-container`

images/server/Containerfile.centos
```diff
@@ -0,0 +1,54 @@
...
+ARG SAMBA_VERSION_SUFFIX=""
+ARG SAMBA_SPECIFICS=daemon_cli_debug_output,ctdb_leader_admin_command
+ARG INSTALL_CUSTOM_REPO=
+
+MAINTAINER John Mulligan <jmulligan@redhat.com>
+
+LABEL org.opencontainers.image.title="Samba container"
+LABEL org.opencontainers.image.description="Samba container"
+LABEL org.opencontainers.image.vendor="Samba in Kubernetes"
+
```
synarete, line 27: Following the same LABELs we use in samba-operrator, please add also org.opencontainers.image.url="https://github.com/samba-in-kubernetes/samba-container

synarete (commented):
LGTM

obnoxxx:
> I'll repeat myself when I recommended that you squash all the patches for this PR together. I think you'll find it much easier to manage and given that the changes to add Centos and use OS specific variables are pretty interelated I don't think it loses much structure or "cleanness" in this case.

@phlogistonjohn  is this request still active? I thought I had squashed according to our discussion, and as much as I coud stand ... 😉

phlogistonjohn:
> > I'll repeat myself when I recommended that you squash all the patches for this PR together. I think you'll find it much easier to manage and given that the changes to add Centos and use OS specific variables are pretty interelated I don't think it loses much structure or "cleanness" in this case.
> 
> @phlogistonjohn is this request still active? I thought I had squashed according to our discussion, and as much as I could stand ... wink

I alredy answered this question:
https://github.com/samba-in-kubernetes/samba-container/pull/107#issuecomment-1508921957

obnoxxx:
@synarete  wrote:
> LGTM

does this comment relate to any review action? 😉

images/server/Containerfile.centos
obnoxxx: @synarete - fwiw, I did this in a separate commit and only changed the centos and fedora based images for a start. opensuse can and probably should be treated separately. what do you think?

Makefile
obnoxxx: @synarete wrote:

> IMHO `print-vars` or `report-vars` if a better target name.

good idea! are you ok with changing this later in a follow-up PR?

Makefile
obnoxxx: @synarete  wrote:
> I prefer using the built-in [info]
(https://www.gnu.org/software/make/manual/make.html#index-info) function for such printing. It is just a matter of style -- no changes need here.

interesting idea again! I will also look at this later, separately. ...

obnoxxx:
fixed one bug in a containerfile. now all builds that were failing in the CI are succeeding locally for me

obnoxxx:
amended the label commit once more to fix the toolbox build

images/server/Containerfile.centos
synarete: Fair enough

synarete (approved):

anoopcs9 (approved):

PR #108, opened 2023-02-13
Makefile organization and cleanups

phlogistonjohn:
Update the Makefile to improve the regularity and DRY-ness of the
container build rules. Each build rule is now made up of common
top-level variables, of which mostly just the names and default build
args change.

I think this improves the readability of the file as well as making it
easier to include the Makefile and then re-use the _img_build target
rather than copy and paste the  build + buildfile creation commands.

---

Add some "section" level header comments to help future
additions/changes to this file be organized. It also reorders a few
of the rules in order to follow a consistent pattern - in particular
every container build rule will be followed by the matching container
push rule.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,19 +65,26 @@ build: build-server build-nightly-server build-ad-server build-client \
 build-server: $(BUILDFILE_SERVER)
 .PHONY: build-server
 $(BUILDFILE_SERVER): Makefile $(SERVER_SRC_FILE) $(SERVER_SOURCES)
-	$(BUILD_CMD) \
-		--tag $(SERVER_NAME) --tag $(SERVER_REPO_NAME) \
-		-f $(SERVER_SRC_FILE) $(SERVER_DIR)
-	$(CONTAINER_CMD) inspect -f '{{.Id}}' $(SERVER_NAME) > $(BUILDFILE_SERVER)
+	$(MAKE) _img_build \
+		BUILD_ARGS=""  \
+		EXTRA_BUILD_ARGS="$(EXTRA_BUILD_ARGS)" \
+		SHORT_NAME=$(SERVER_NAME) \
+		REPO_NAME=$(SERVER_REPO_NAME) \
+		SRC_FILE=$(SERVER_SRC_FILE) \
+		DIR=$(SERVER_DIR) \
+		BUILDFILE=$(BUILDFILE_SERVER)
 
 build-nightly-server: $(BUILDFILE_NIGHTLY_SERVER)
 .PHONY: build-nightly-server
 $(BUILDFILE_NIGHTLY_SERVER): Makefile $(SERVER_SRC_FILE) $(SERVER_SOURCES)
-	$(BUILD_CMD) \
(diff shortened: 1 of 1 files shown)


anoopcs9 (approved):
Good cleanup, thanks.

Makefile
```diff
@@ -154,3 +182,26 @@ check-shell-scripts:
...
+_img_build:
+	$(BUILD_CMD) \
+		$(BUILD_ARGS) \
+		$(EXTRA_BUILD_ARGS) \
+		--tag $(SHORT_NAME) \
+		--tag $(REPO_NAME) \
+		-f $(SRC_FILE) \
+		$(DIR)
+	$(CONTAINER_CMD) inspect -f '{{.Id}}' $(SHORT_NAME) > $(BUILDFILE)
+.PHONY: _img_build
```
synarete, line 217: This is a valid solution. Personally, I prefer using [call-function](https://www.gnu.org/software/make/manual/html_node/Call-Function.html) instead of invoking sub-make rule. Something like:

```
define _img_build2
	$(BUILD_CMD) $(1) $(2) --tag $(3) --tag $(4) -f $(5) $(6)
	$(CONTAINER_CMD) inspect -f '{{.Id}}' $(3) > $(7)
endef
```

And use it with:

```
build-toolbox: $(BUILDFILE_TOOLBOX)
.PHONY: build-toolbox
$(BUILDFILE_TOOLBOX): Makefile $(TOOLBOX_SRC_FILE)
	$(call _img_build2, , $(EXTRA_BUILD_ARGS), \
		$(TOOLBOX_NAME), $(TOOLBOX_REPO_NAME), \
		$(TOOLBOX_SRC_FILE), $(TOOLBOX_DIR), $(BUILDFILE_TOOLBOX) )
```

I assume its a matter of personal preference. Both solutions are valid.

Makefile
```diff
@@ -65,19 +65,26 @@ build: build-server build-nightly-server build-ad-server build-client \
 build-server: $(BUILDFILE_SERVER)
 .PHONY: build-server
 $(BUILDFILE_SERVER): Makefile $(SERVER_SRC_FILE) $(SERVER_SOURCES)
-	$(BUILD_CMD) \
-		--tag $(SERVER_NAME) --tag $(SERVER_REPO_NAME) \
-		-f $(SERVER_SRC_FILE) $(SERVER_DIR)
-	$(CONTAINER_CMD) inspect -f '{{.Id}}' $(SERVER_NAME) > $(BUILDFILE_SERVER)
+	$(MAKE) _img_build \
```
synarete, line 71: When invoking sub-make, it is always a good idea to call it with explicit directory (`$(MAKE) -C $(DIR)`). After all, someone might call top-level `Makefile` from another directory (something like `make -f samba-container/Makefile`).

synarete (approved):
LGTM

Makefile
phlogistonjohn: I went looking for how to do that but the gnu make docs are a bit meh in my opinion and I wasn't fully up on functions in make. The other issue I have with call-function is that I much prefer named to positional parameters and since I didn't see a way to do named params with a function I reverted to the approach I knew best.

PR #109, opened 2023-02-17
makefile: use consistent naming

phlogistonjohn:
Previously, we had local image names like "samba-container" or
"samba-client-container" while the repos (sans prefix) were name like
"samba-server" or "samba-client". This makes it more difficult to match
up the images and reuse the variables.
This patch updates the local image names to match the public ones:
* samba-container becomes samba-server
* samba-ad-container becomes samba-ad-server
* samba-client-container becomes samba-client
... and so on.

---

In the previous commit we made local image names match the ones on the
global container registry. This change updates the opensuse specific makefile to
also have local names that match repo names. Examples:
* samba-opensuse-container becomes samba-server
* samba-client-opensuse-container becomes samba-client
... and so on.


obnoxxx (approved):
lgtm, thanks

phlogistonjohn:
cc: @dmulder   I made some changes to the opensuse build file to make them consistent with the main Makefile, please take a look. I don't think there should be an issue because it would only impact the local names but I still want to make sure you get a chance to look at it.

obnoxxx (approved):
all checks passing now. still lgtm

anoopcs9 (approved):
lgtm, thanks.

anoopcs9:
> cc: @dmulder I made some changes to the opensuse build file to make them consistent with the main Makefile, please take a look. I don't think there should be an issue because it would only impact the local names but I still want to make sure you get a chance to look at it.

Oops ! I thought you might have set some label to avoid automatic merge until @dmulder gets a chance to look at the changes...sorry.

phlogistonjohn:
> > cc: @dmulder I made some changes to the opensuse build file to make them consistent with the main Makefile, please take a look. I don't think there should be an issue because it would only impact the local names but I still want to make sure you get a chance to look at it.
> 
> Oops ! I thought you might have set some label to avoid automatic merge until @dmulder gets a chance to look at the changes...sorry.

I probably should have.
@dmulder if you see anything here that you have questions or concerns about, please don't hesitate to report it in an issue.

Makefile.opensuse
```diff
@@ -5,10 +5,10 @@ AD_SERVER_SRC_FILE:=$(AD_SERVER_DIR)/Containerfile.opensuse
...
 TOOLBOX_SRC_FILE:=$(TOOLBOX_DIR)/Containerfile.opensuse
 SERVER_SOURCES:=$(SERVER_DIR)/smb.conf
-SERVER_NAME:=samba-opensuse-container:$(TAG)
-AD_SERVER_NAME:=samba-ad-opensuse-container:$(TAG)
-CLIENT_NAME:=samba-client-opensuse-container:$(TAG)
-TOOLBOX_NAME:=samba-toolbox-opensuse-container:$(TAG)
+SERVER_NAME:=samba-server:$(TAG)
+AD_SERVER_NAME:=samba-ad-server:$(TAG)
+CLIENT_NAME:=samba-client:$(TAG)
+TOOLBOX_NAME:=samba-toolbox:$(TAG)
```
dmulder, line 11: This will now overwrite the fedora images (notice you removed '-opensuse' from each of the names). Was this intentional?

phlogistonjohn: Yes, the "unqualified" or "local names" were inconsistent with the fully qualified (or repo) names. @obnoxxx is working on changes to make building on various bases more organized and consistent and one of the stumbling blocks we had was variable reuse was hard due to these naming differences.

We made this change together as a building block for that. So yes, right now if you run `make build-server` and following run `make -f Makefile.opensuse build-server` the short name will be overwritten. I think that OK and may even improve our ability to reuse the tests for various base images.

dmulder: Ok, you don't need these entries at all in Makefile.opensuse then. Now all we're doing is overridding the existing names with the same names.

PR #115, opened 2023-04-04
Makefile: improve buildfile handling, including cleanup

obnoxxx:
This change improves the definitions of buildfile variables by introducing a variable for the filename prefix, '.build'.

As a side effect, the clean target which previously just deleted a somewhat arbitrary explicit list of buildfiles and was therefore error-prone, could be
made more systematic and robust by introducing the use of a wildcard. This way, the clean target should continue working even if future changes change the exact names of buildfiles.

This change was separated out of PR #107  where it was a mere byproduct

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,12 +54,13 @@ NIGHTLY_AD_SERVER_REPO_NAME:=quay.io/samba.org/samba-ad-server:nightly
 CLIENT_REPO_NAME:=quay.io/samba.org/samba-client:$(TAG)
 TOOLBOX_REPO_NAME:=quay.io/samba.org/samba-toolbox:$(TAG)
 
-BUILDFILE_SERVER:=.build.server
-BUILDFILE_NIGHTLY_SERVER:=.build.nightly-server
-BUILDFILE_AD_SERVER:=.build.ad-server
-BUILDFILE_NIGHTLY_AD_SERVER:=.build.nightly-ad-server
-BUILDFILE_CLIENT:=.build.client
-BUILDFILE_TOOLBOX:=.build.toolbox
+BUILDFILE_PREFIX=.build
+BUILDFILE_SERVER:=$(BUILDFILE_PREFIX).server
+BUILDFILE_NIGHTLY_SERVER:=$(BUILDFILE_PREFIX).nightly-server
+BUILDFILE_AD_SERVER:=$(BUILDFILE_PREFIX).ad-server
+BUILDFILE_NIGHTLY_AD_SERVER:=$(BUILDFILE_PREFIX).nightly-ad-server
+BUILDFILE_CLIENT:=$(BUILDFILE_PREFIX).client
+BUILDFILE_TOOLBOX:=$(BUILDFILE_PREFIX).toolbox
 
 build: build-server build-nightly-server build-ad-server build-client \
 	build-toolbox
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -54,12 +54,13 @@ NIGHTLY_AD_SERVER_REPO_NAME:=quay.io/samba.org/samba-ad-server:nightly
...
-BUILDFILE_NIGHTLY_AD_SERVER:=.build.nightly-ad-server
-BUILDFILE_CLIENT:=.build.client
-BUILDFILE_TOOLBOX:=.build.toolbox
+BUILDFILE_PREFIX=.build
+BUILDFILE_SERVER:=$(BUILDFILE_PREFIX).server
+BUILDFILE_NIGHTLY_SERVER:=$(BUILDFILE_PREFIX).nightly-server
+BUILDFILE_AD_SERVER:=$(BUILDFILE_PREFIX).ad-server
+BUILDFILE_NIGHTLY_AD_SERVER:=$(BUILDFILE_PREFIX).nightly-ad-server
+BUILDFILE_CLIENT:=$(BUILDFILE_PREFIX).client
+BUILDFILE_TOOLBOX:=$(BUILDFILE_PREFIX).toolbox
```
synarete, line 63: Cool!

obnoxxx:
the test-nightly-ad-derver ci job failed the same way as seen in other prs and regular runs. I think it's not related to this PR

Makefile
```diff
@@ -193,7 +194,7 @@ check-shell-scripts:
 ### Mics. Rules ###
 
 clean:
-	$(RM) $(BUILDFILE_SERVER) $(BUILDFILE_NIGHTLY_SERVER) $(BUILDFILE_AD_SERVER) $(BUILDFILE_NIGHTLY_AD_SERVER) $(BUILDFILE_CLIENT) $(BUILDFILE_TOOLBOX)
+	$(RM) $(BUILDFILE_PREFIX)*
```
synarete, line 197: This is ok, but as a general rule I prefer explicit `$(RM)` in makefiles. Using wildcard may have some unexpected side-effects; see [Wildcard-Pitfall](https://www.gnu.org/software/make/manual/html_node/Wildcard-Pitfall.html).

For now, lets approve this and see in the future if it needs refinements.

synarete (approved):
LGTM

Makefile
obnoxxx: @synarete thanks for your review! It is correct that wildcards can bring pitfalls with them  and often, using explicit names instead can be more robust, but as it happened, in my work on PR #107 , the pitfall/problem was due to the fact that the clean target did *not* use a wildcard .

phlogistonjohn:
`kefile`?  Oh, it's `makefile` minus the `ma`. Please fix the commit message (ideally the pr title too, but I care less about that because PRs are less permanent).

Makefile
phlogistonjohn: I'm OK with this for now. If it causes problems we can revisit it in the future in a more explicit way.

obnoxxx:
> `kefile`? Oh, it's `makefile` minus the `ma`. Please fix the commit message (ideally the pr title too, but I care less about that because PRs are less permanent).

done, fixed both

anoopcs9 (approved):

phlogistonjohn (approved):

PR #114, opened 2023-03-27
add tdb-tools to ad-server and toolbox images

phlogistonjohn:
Including tdb tools will permit the use of programs like tdbbackup
to be used from within the AD DC container.

Fixes: #113

----

Also add tdb-tools to the toolbox images, since they are tools.

PR #117, opened 2023-04-12
drop jsonschema rpms

phlogistonjohn:
Temporarily remove jsonschema from our packages list. sambacc tries to
    use this package but is not compatible with the version in fedora 36

### DIFF
--- a/images/ad-server/install-packages.sh
+++ b/images/ad-server/install-packages.sh
@@ -27,7 +27,6 @@ esac
 dnf install --setopt=install_weak_deps=False -y \
     findutils \
     python-pip \
-    python3-jsonschema \
     python3-samba \
     python3-pyxattr \
     "samba-dc${samba_version_suffix}" \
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
I think this is the last of the needed jsonschema releated fixes. I think the test failure for test-nightly-ad-server-kubernetes  is unrelated and is the failure we were seeing previously, and thus can be ignored for now.

anoopcs9 (approved):

synarete (approved):
LGTM

anoopcs9:
> Temporarily remove jsonschema from our packages list. sambacc tries to use this package but is not compatible with the version in fedora 36

As mentioned in https://github.com/samba-in-kubernetes/samba-container/pull/114#issuecomment-1497642231, we now have 2 options to satisfy the dependency on _python3-markdown_ for nightly images:

[1] Use the current patch to install _python3-markdown_ for ad-server images.
[2] Add `Requires:` for _python3-markdown_ in RPM spec file.

anoopcs9:
> [2] Add `Requires:` for _python3-markdown_ in RPM spec file.

PR https://github.com/samba-in-kubernetes/samba-build/pull/4

PR #119, opened 2023-04-18
ad-server: add tdb-tools to installed packages

dmulder:
Including tdb tools will permit the use of programs like tdbbackup to be used from within the AD DC container.

This was added to the Fedora container, so let's add it to the openSUSE container for consistency.


dmulder:
The CI failure has nothing to do with this change.

anoopcs9 (approved):

obnoxxx (approved):
igtm

obnoxxx:
@mergifyio rebase

Issue #121, opened 2023-04-20
add gitlint commit linting


obnoxxx:
similar to https://github.com/samba-in-kubernetes/samba-operator/pull/298

PR #124, opened 2023-04-20
Add gitlint support to samba-container

phlogistonjohn:
Fixes: #121

I'm perfectly happy with the first commit in this series. 

Everything including and after adding the 'install-tools.sh' I'm a bit iffy on. I tried multiple approaches and nothing pleased me. This current version is based on the workflow we have in samba-container where if the makefile determines a prerequisite tool is not present it tries to locally install one in the `.bin` subdir under the repo root.  Unlike the samba-operator version I tried really hard to avoid adding a lot of boilerplate to the makefie now and in the future and used some pattern matching to avoid it. Still, I'm not exactly thrilled with the implementation, but it was working OK for me.

### DIFF
--- a/hack/install-tools.sh
+++ b/hack/install-tools.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+# helper script to install build auxiliary tools in local directory
+#
+# usage:
+#   install-tools.sh --<tool-name> <alt_bin>
+#
+#
+set -e
+
+_ensure_alt_bin() {
+	mkdir -p "${ALT_BIN}"
+}
+
+_require_py() {
+    if [ -z "$PY_CMD" ]; then
+        echo "error: python3 command required, but not found" >&2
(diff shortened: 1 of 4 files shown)


hack/install-tools.sh
```diff
@@ -0,0 +1,77 @@
...
+            exit 0
+        fi
+        _install_shellcheck 1>&2
+        echo "${installed_to}"
+    ;;
+    *)
+        echo "usage: $0 --<tool-name> [<ALT_BIN>]"
+        echo ""
+        echo "available tools:"
+        echo "  --gitlint"
```
anoopcs9, line 75: Did you miss `shellcheck` here?

anoopcs9 (changes requested):
All good. See below for a minor request.

anoopcs9 (approved):
lgtm.

obnoxxx (approved):
lgtm, thanks for doing this!

Issue #95, opened 2023-01-17
Including additional VFS modules


phlogistonjohn:
Currently when we install samba we leave out certain VFS modules. These are modules that are packaged separately in fedora:
samba-vfs-cephfs
samba-vfs-glusterfs
samba-vfs-iouring

Should we just include them in all container images? Or would it be better to have some sort of alternate image/layer to handle it?

phlogistonjohn:
Note: we don't have to handle them all the same. We could choose to always include iouring but have additional layers for cephfs and/or glusterfs (for example).

PR #127, opened 2023-04-25
server: install some vfs packages

phlogistonjohn:
Install the cephfs and iouring vfs packages. These vfs modules are part of samba but are not packaged with the base server. Add cephfs so we can experiment with it soon.
Add iouring so we can experiment with it later.

Fixes: #95

### DIFF
--- a/images/server/install-packages.sh
+++ b/images/server/install-packages.sh
@@ -31,10 +31,10 @@ esac
 
 dnf_cmd=(dnf)
 if [[ "${OS_BASE}" = centos ]]; then
-  dnf_cmd+=(--enablerepo=crb --enablerepo=resilientstorage)
+    dnf_cmd+=(--enablerepo=crb --enablerepo=resilientstorage)
 fi
-"${dnf_cmd[@]}" \
-    install --setopt=install_weak_deps=False -y \
+
+packages=(\
     findutils \
     python-pip \
     python3-samba \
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
It appears that either the vfs modules in question are not built on centos or are in other repos that we don't have enabled. @anoopcs9 do you know?

phlogistonjohn:
Made the script include the vfs-cephfs package only on fedora for now.

anoopcs9:
> It appears that either the vfs modules in question are not built on centos or are in other repos that we don't have enabled. @anoopcs9 do you know?

I thought I replied here. Anyway this is similar to the other case where AD DC components are not available with CentOS. I can try including it(along with AD DC compoents) in nightly builds for CentOS Stream 9 in future.

images/server/install-packages.sh
```diff
@@ -43,8 +39,18 @@ fi
     "samba-client${samba_version_suffix}" \
     "samba-winbind${samba_version_suffix}" \
     "samba-winbind-clients${samba_version_suffix}" \
+    "samba-vfs-iouring${samba_version_suffix}" \
     tdb-tools \
-    "ctdb${samba_version_suffix}"
+    "ctdb${samba_version_suffix}")
+if [[ "${OS_BASE}" = centos ]]; then
+    dnf_cmd+=(--enablerepo=crb --enablerepo=resilientstorage)
+fi
```
anoopcs9, line 47: Purely a cosmetic change request but the above block can still stay after LINENO:32.

anoopcs9 (approved):
lgtm with a minor comment.

obnoxxx (commented):
The change looks good to me overall, but I noticed that the mentioned issue #95 mentions the glusterfs module, too. Should we install it?

images/server/install-packages.sh
phlogistonjohn: done

phlogistonjohn:
> The change looks good to me overall, but I noticed that the mentioned issue #95 mentions the glusterfs module, too. Should we install it?

done

gd (commented):
LGTM, RB+

gd (commented):
2nd try... LGTM, RB+

obnoxxx (approved):
lgtm

PR #131, opened 2023-05-15
Add preliminary support for building arm64 images

phlogistonjohn:
Aside from a couple of cleanup commits this series contains two significant changes:
1. It adds a new form of image name/tag I've nicknamed the "fully qualified image name" that is a long tag containing a bunch of details about the image - mainly package source (default or nightly), base os, and architecture. Plus developers can add a final optional extra value when hacking on things.
2. Add a BUILD_ARCH variable that you can pass to make to have it build an architecture other than the native one.

use it like `make BUILD_ARCH=arm64` (assuming an amd64 machine).

Note that this uses a standard way to request arm64 (or other) architectures using     the `--arch` flag to podman (untested on docker (yet)). This makefile change does *nothing* to     attempt to make that flag work on a particular system. If the flag     works then this makefile will build the image and save it, with
    the correct arch recorded in the recently added "fully qualified     image name".

This change also does nothing with with image, no tests, no creating     of manifest lists to create a something pushable, nothing.     I tried to start some of that, which lead to me attempting to lightly
    reorganize the makefile, which lead me to attempting to heavily     reorganize the makefile, and that was a huge mess and a terrible slog     and the delta would have been impossible to follow. Thus I'm now     creating the smallest delta I can think of that still gets the work     to support arm64 builds started.

Future PRs will attempt to further use the FQIN to disambiguate the image builds so that building images won't be so much based on variables and buildfiles that overlap. As well as beginning the work to make this testable(?) and pushable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,32 @@ BUILDFILE_NIGHTLY_CLIENT=$(BUILDFILE_PREFIX).$(OS_PREFIX)nightly-client
 BUILDFILE_TOOLBOX=$(BUILDFILE_PREFIX).$(OS_PREFIX)toolbox
 BUILDFILE_NIGHTLY_TOOLBOX=$(BUILDFILE_PREFIX).$(OS_PREFIX)nightly-toolbox
 
+HOST_ARCH:=$(shell arch)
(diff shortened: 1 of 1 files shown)


anoopcs9 (approved):
I don't know if something else is better but generally it looks good.

PR #133, opened 2023-05-23
[WIP][RFC] use fqin in buildfiles to create unique images

phlogistonjohn:
Just seeing if this will pass CI, this is messy and not easily reviewable at the moment, but you can take a peek :-)

Depends on: #131

PR #135, opened 2023-05-31
[WIP][RFC] build images using new hack/build-image script

phlogistonjohn:
Conflicts with #133 

This PR is a working prototype of an alternate approach to #133. In this patch series I remove a lot of the complexity from the makefile and add a new script `build-image` written in python to the hack dir.

While the python ends up being longer, it allows the use of real data structures and is (IMO) a lot more straightforward and easy to read than the makefile based changes in #133.

I want the team to take a look at the overall diff of both and we can debate which one is the better overall approach... or just take both as ideas and suggest a 3rd approach. ;-)


obnoxxx:
@mergifyio rebase

obnoxxx (dismissed):
Thanks for doing this, @phlogistonjohn  !

This is a clean improvement of the build mechanisms.

I reviewed the commits, read the new script, and played with it locally.

LGTM.

obnoxxx:
@phlogistonjohn needs fixes for failing checks now after rebase

phlogistonjohn:
Seems like that job might be flaky. This is the 2nd time I just reran it and it worked on the retry.

Makefile
```diff
@@ -128,34 +56,18 @@ build: build-server build-nightly-server build-ad-server build-client \
...
-	@echo NIGHTLY_TOOLBOX_NAME: $(NIGHTLY_TOOLBOX_NAME)
-	@echo NIGHTLY_TOOLBOX_REPO_NAME: $(NIGHTLY_TOOLBOX_REPO_NAME)
 
 	@echo BUILDFILE_SERVER: $(BUILDFILE_SERVER)
 	@echo BUILDFILE_AD_SERVER: $(BUILDFILE_AD_SERVER)
 	@echo BUILDFILE_NIGHTLY_AD_SERVER: $(BUILDFILE_iNIGHTLY_AD_SERVER)
 	@echo BUILDFILE_NIGHTLY_SERVER: $(BUILDFILE_NIGHTLY_SERVER)
 	@echo BUILDFILE_CLIENT: $(BUILDFILE_CLIENT)
 	@echo BUILDFILE_TOOLBOX: $(BUILDFILE_TOOLBOX)
-	@echo BUILDFILE_NIGHTLY_TOOLBOX: $(BUILDFILE_NIGHTLY_TOOLBOX)
```
anoopcs9, line 158: What was the rationale behind removing specific lines and not others?

anoopcs9 (commented):
Why do we have the new Python script named _build-image_ and not _build-image.py_?

phlogistonjohn:
> Why do we have the new Python script named build-image and not build-image.py?

First, I wanted to ensure that if we ever decided to change how the tool is implemented we wouldn't have to rename it. Also, it's also very common  to name executables in python without the .py extension (see commands like `black`, `yamllint`, `hg`, `copr-cli`,  `dnf`,  `ansible-playbook` and the list goes on). Only python files that expected to be imported need to have a py extension. So I think it's the more typical naming approach for an executable (albeit one that isn't expected to be installed).

Makefile
phlogistonjohn: I tried to remove variables that were no longer useful/interesting. BUILDFILE_NIGHTLY_TOOLBOX was a special case as it was never used anywhere else in the file.  I expect to have missed some stuff in my cleanups, the Makefile was (and still is) pretty messy. In the future we may be able to drop this debug target entirely, but I am trying not to make this PR too big and drastic.

Makefile
```diff
@@ -47,72 +30,17 @@ AD_SERVER_SOURCES=\
...
-# without the tag part.
-# Usage: $(call get_imagename,<image-name>)
-get_imagename=$(firstword $(subst :, ,$1))
-
-# get_pkgsource is a function macro that, given an images name returns
-# the name of the package source. Currently only understands the
-# difference between default (os packages) and nightly (SIT packages).
-# Usage: $(call, get_pkgsource,<image-name>)
-get_pkgsource=$(if $(findstring nightly,$1),nightly,default)
+_BUILD_KP=$(BUILD_IMAGE) $(if $(CONTAINER_CMD),--container-engine=$(CONTAINER_CMD)) $(BI_PREFIX_ARGS) --kind=$1 --package-source=$2 --distro-base=$(SRC_OS_NAME)  --repo-base=$(REPO_BASE) $(if $(BUILD_ARCH),--arch=$(BUILD_ARCH)) $3
```
synarete, line 43: Line too long. Consider split into multi-lines

hack/build-image
```diff
@@ -0,0 +1,597 @@
...
+
+_DISCOVERED_CONTAINER_ENGINES = []
+
+
+def check_kind(kind):
+    """Return the canonical name for the image kind or raise a ValueError."""
+    try:
+        return IMAGE_KINDS[kind]
+    except KeyError:
+        raise ValueError(f"invalid kind: {kind}")
```
synarete, line 121: Consider:
```
except KeyError as ex:
    raise ValueError(f"invalid kind: {kind}") from ex
```

hack/build-image
```diff
@@ -0,0 +1,597 @@
...
+
+
+def check_arch(arch):
+    """Return the canonical name for the arch or raise a ValueError."""
+    try:
+        return ARCHITECTURES[arch]
+    except KeyError:
+        raise ValueError(f"invalid arch: {arch}")
```
synarete, line 129: Consider:
```
except KeyError as ex:
    raise ValueError(f"invalid arch: {arch}") from ex
```

hack/build-image
```diff
@@ -0,0 +1,597 @@
...
+def _cmd_to_str(cmd):
+    """Format a command for logging."""
+    return " ".join(shlex.quote(arg) for arg in cmd)
+
+
+def run(cli, cmd, capture_output=False, check=False):
+    """Execute a command. Wraps subprocess.run."""
+    if cli.dry_run and not capture_output:
+        logger.info("Would run: %s", _cmd_to_str(cmd))
+        return
```
synarete, line 166: Mixed return type int this function. Better have `return ""` here for future type-hints annotations.

hack/build-image
```diff
@@ -0,0 +1,597 @@
...
+    return subprocess.run(cmd, capture_output=capture_output, check=check)
+
+
+def container_engine(cli):
+    """Return the path to a container engine. If the container engine is not
+    yet known, discover it and cache the result.
+    """
+    eng = cli.container_engine
+    if eng:
+        logger.info(f"Using specified container engine: {eng}")
```
synarete, line 177: Other logging use formatted output, so better have it here has well: `logger.info("Using specified container engine: %s", eng)`

hack/build-image
```diff
@@ -0,0 +1,597 @@
...
+    def __init__(self, default_repo_base, distro_repo=None):
+        self.default = default_repo_base
+        self.distro_map = dict(distro_repo or [])
+
+    def find_base(self, distro):
+        try:
+            return self.distro_map[distro]
+        except KeyError:
+            pass
+        return self.default
```
synarete, line 264: Better have: `return self.distro_map.get(distro, self.default)`

hack/build-image
```diff
@@ -0,0 +1,597 @@
...
+    container_push(cli, push_name)
+
+
+def print_buildfile(cli, target):
+    """Command to print build file names."""
+    build_file = pathlib.Path(f"{cli.buildfile_prefix}{target.flat_name()}")
+    print(build_file)
+
+
+def print_image(cli, target):
```
synarete, line 417: Unused argument: `cli`

hack/build-image
```diff
@@ -0,0 +1,597 @@
...
+    """Command to print (fqin) image names."""
+    print(str(target))
+
+
+def print_tags(cli, target, with_tags=True):
```
synarete, line 422: Unused argument: `with_tags`

images/toolbox/Containerfile.centos
```diff
@@ -1,3 +1,8 @@
+# FIXME - this is not a real tag publicly available in the
+# quay.io/samba.org/samba-client repository. This only works if you build
+# the centos client locally first or acquire the image from a side channel.
+# This needs to be converted to something public and/or configurable
+# later.
```
synarete, line 5: This raise two questions:
1) What is the appropriate tag we should use?
2) Should we still use `samba-client` as base image for `toolbox` or maybe use other (centos,fedora,opensuse) as base?

synarete (commented):
Using python script improves code quality and readability, but see review comments.

Makefile
phlogistonjohn: It's a good idea but make can be weirdly sensitive to line continuations with `\` .  If you don't mind I'd like to leave it as-is for now. We can always come back to it in a future patch. Is that OK?

hack/build-image
phlogistonjohn: Exception chaining can be handy when you want to retain the info from the inner exception, but here, the fact that we use a dict lookup to validate the value feels like an implementation detail to me. Do you feel strongly about this?

hack/build-image
phlogistonjohn: Will update.

hack/build-image
phlogistonjohn: Will update

hack/build-image
phlogistonjohn: OK

hack/build-image
phlogistonjohn: Will update

Makefile
synarete: Sure; it is just an idea.

images/toolbox/Containerfile.centos
phlogistonjohn: > This raise two questions:
> 
>     1. What is the appropriate tag we should use?
> 

I don't know all I know is this will break in some circumstances. But I didn't want this PR to become sidetracked by that issue so I didn't try to fix it, just left a breadcrumb to follow up on later.

>     2. Should we still use `samba-client` as base image for `toolbox` or maybe use other (centos,fedora,opensuse) as base?

I think layering the images is fine, but there are some downsides to it. Let's discuss this more in a new issue or meeting, etc.  :-)

hack/build-image
synarete: No; it is more a common habit I carry with me. As said in [https://legacy.python.org/dev/peps/pep-3134/](PEP-3134), "In order to debug the problem, it is useful to know about both exceptions" but I guess that here it is just a "nice to have".

hack/build-image
phlogistonjohn: fixed

obnoxxx (approved):
still LGTM. :-)

PR #136, opened 2023-06-07
Workflow matrix

obnoxxx:
convert workflow to matrix to test OS_NAME vs BUILD_ARCH and add opensuse to the matrix


.github/workflows/container-image.yml
```diff
@@ -37,72 +37,103 @@ jobs:
...
+      os: [centos, fedora, opensuse]
+      arch: [amd64, x86_64]
+
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - name: Build the centos client image
-      run: make OS_NAME=centos build-client
+    -  name: build the client image 
+    -  run: make OS_NAME=${{ matrix.os }} BUILD_ARCH=${{ matrix.arch }} build-client
```
anoopcs9, line 121:
```suggestion
    - name: build the client image 
      run: make OS_NAME=${{ matrix.os }} BUILD_ARCH=${{ matrix.arch }} build-client
```

.github/workflows/container-image.yml
```diff
@@ -37,176 +35,174 @@ jobs:
       - name: Lint git commit messages
         run: make check-gitlint
 
+
+
+
+    strategy:
+      matrix:
+        os: [centos, fedora]
+        arch: [amd64, x86_64]
```
phlogistonjohn, line 44: I don't know if you meant to do this as an experiment or it's a mistake. `amd64` and `x86_64` are the same architecture. One's the "golang" style of name and the other is the "kernel" style of name. I have no idea if trying `arm64` works on the CI infra yet, but I was wondering if you meant to put it in the list.

obnoxxx:
`unknown flag: --arch` strange! ...

obnoxxx:
the remaining problem seems to be the lack of podman support of the docker-image-artifact action: https://github.com/ishworkh/docker-image-artifact/issues/3 @phlogistonjohn - @anoopcs9

obnoxxx:
@phlogistonjohn, I rebased on top of master after #138  merged, but test still fail with cancelled operations ...

obnoxxx:
trying to run `make test-server` locally, I get a different error:
```
$ make test-server
Error: lsetxattr /var/folders/hs/vmf05zhn49b22zvpjxpc0gm80000gn/T/tmp.lj2dCA6s/containers-user-501: operation not supported
Error running samba container
$ 
```
this might be related to podman on MacOS ...

obnoxxx:
@mergifyio rebase

.github/workflows/container-image.yml
```diff
@@ -37,250 +35,227 @@ jobs:
...
-    - name: Build the nightly server image
-      run: make build-nightly-server
-    - name: Upload nightly server image
-      uses: ishworkh/docker-image-artifact-upload@v1
-      with:
-        image: "samba-server:nightly"
-        retention_days: 1
+    strategy:
+      matrix:
+         os: [centos, fedora, opensuse]
```
phlogistonjohn, line 146:
```suggestion
         os: [centos, fedora]
```

.github/workflows/container-image.yml
```diff
@@ -37,250 +35,227 @@ jobs:
...
-    - name: Test the nightly server image
-      run: LOCAL_TAG=samba-server:nightly tests/test-samba-container.sh
 
-  test-nightly-server-centos:
-    needs: build-nightly-server-centos
+
+  test-nightly-servers:
+    strategy:
+      matrix:
+         os: [centos, fedora, opensuse]
```
phlogistonjohn, line 176:
```suggestion
         os: [centos, fedora]
```

.github/workflows/container-image.yml
```diff
@@ -37,250 +35,227 @@ jobs:
...
     needs:
       - build-ad-server
       - build-server
-    #runs-on: ubuntu-latest
     # need to explicitly use 20.04 to avoid problems with jq...
     runs-on: ubuntu-20.04
-    steps:
+env:
+  IMG_TAG: ${{ matrix.os }}-latest
+steps:
```
phlogistonjohn, line 208:
```suggestion
    env:
      IMG_TAG: ${{ matrix.os }}-latest
    steps:
```

.github/workflows/container-image.yml
```diff
@@ -37,250 +35,227 @@ jobs:
...
-        - build-nightly-server
-      #runs-on: ubuntu-latest
-      # need to explicitly use 20.04 to avoid problems with jq...
-      runs-on: ubuntu-20.04
-      env:
-        IMG_TAG: nightly
-      steps:
+    strategy:
+      matrix:
+         os: [centos, fedora, opensuse]
```
phlogistonjohn, line 230:
```suggestion
         os: [centos, fedora]
```

phlogistonjohn:
Hi @obnoxxx, as we discussed earlier today I have pushed some additional patches into your PR. All the jobs that ran are green at the moment. However, there are still a number of issues with the PR:
* There's excess whitespace in the YAML file.
* The changes to the structure of the actions trips up github's "required statuses" and it acts like some statuses have not been reported. This is despite some of them having the same name.
* Due to the above, I don't think this PR can be automerged, once approved, we'll probably need to fix mergify rules and merge it manually
* I'm not entirely sure how the `push` target will work if we merge these matrix changes. After reading the code a few times I *think* we are probably OK and the `push` job will just rebuild everything it needs, but I am not certain.
* There are a bunch of "fixup" patches in the PR. I think I'd prefer to squash all the changes to container-image.yml into a single patch.

Now that I've commited directly to this PR, I am willing to do more, but I also want to give you a chance to make changes as needed. First, let's deal with the witespace issue. Let me know if you want me to push a change to clean up the (new) whitespace issues in the yaml file.  After that, we can squash things and then get reviews before dealing with the other things.

phlogistonjohn:
Oh, and you'll probably want to look at my current changes and provide feedback on those too. :-)

obnoxxx:
@phlogistonjohn  wrote:

> Oh, and you'll probably want to look at my current changes and provide feedback on those too. :-)

The additional patches look good to me, thanks!

I don't see the whitespace problem in the yaml yet. feel free to fix it in the PR directly .

synarete (commented):
Commit message has a type: "thetest" instead of "the test".
Plus, there is redundant white space before "test-server".

Consider using `vim` as you git-commit editor and add `autocmd FileType gitcommit setlocal spell` to your `.vimrc`. This will catch those kind of errors upon commit.

tests/test-samba-container.sh
```diff
@@ -1,43 +1,71 @@
...
 		echo "Error running samba container"
-	else
+		exit 1
+	fi
 		echo "Container started, ID: '${CONTAINER_ID}'"
 
 		# give samba a second to come up
 		sleep 1
 
+		echo "Listing samba shares"
```
synarete, line 40: Wrong indentation: should be one-tab less (including the 3 lines above and the following block).

tests/test-samba-container.sh
```diff
@@ -1,43 +1,71 @@
...
 		fi
-	fi
+
+
+echo "stopping samba container."
 
 	${CONTAINER_CMD} kill "${CONTAINER_ID}"
+	rc=$?
+
+	if [ $rc -ne 0 ]; then
```
synarete, line 55: Wrong indentation (extra tab)

.github/workflows/container-image.yml
```diff
@@ -39,202 +37,155 @@ jobs:
...
     - uses: actions/checkout@v3
-    - name: Build the client image
-      run: make build-client
+    - name: build the client image
+      run: make OS_NAME=${{ matrix.os }} BUILD_ARCH=${{ matrix.arch }} build-client
     # Here we upload samba-client image to artifacts locally for consumption
-    # during samba-toolbox build process.
-    - name: Upload client image
+    # during  the samba-toolbox build process.
+    - name: Upload ithe client image
```
synarete, line 88: Typo: ithe --> the

synarete (commented):
Overall looks fine, with just few minor fixes.

tests/test-samba-container.sh
obnoxxx: fixed, thanks!

tests/test-samba-container.sh
obnoxxx: fixed, thanks!

.github/workflows/container-image.yml
obnoxxx: fixed, thanks!

obnoxxx:
@synarete  wrote:
> Commit message has a type: "thetest" instead of "the test". Plus, there is redundant white space before "test-server".
> 

thanks, I fixed those. 

> Consider using `vim` as you git-commit editor and add `autocmd FileType gitcommit setlocal spell` to your `.vimrc`. This will catch those kind of errors upon commit.

thanks for the hint. I might try that.

.github/workflows/container-image.yml
obnoxxx: fixed, thanks!

.github/workflows/container-image.yml
```diff
@@ -39,202 +37,155 @@ jobs:
...
-
-  build-nightly-server-centos:
+  build-nightly-servers:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [centos, fedora]
+        arch: [amd64]
+        exclude:
+          - os: centos
```
anoopcs9, line 137: Why do we exclude `centos` here? Nightly builds works well with CentOS in building server images. I think we should have a common `build-nightly-server` job for `centos` and `fedora` without any exclusion. We should also have a common `build-nightly-ad-server` job for `centos` and `fedora` excluding `centos` similar to `build-ad-server` job.

tests/test-samba-container.sh
```diff
@@ -1,44 +1,71 @@
...
 
+	echo "Listing samba shares"
 		${CONTAINER_CMD} exec "${CONTAINER_ID}" smbclient -U% -L 127.0.0.1
 		rc=$?
 
-		if [ $rc -ne 0 ]; then
+		if [ ${rc} -ne 0 ]; then
 			echo "Error listing samba shares"
+			exit 1
 		fi
```
anoopcs9, line 47: We could remove extra indentation(`tab`) starting from LINENO:25 through LINENO:47.

.github/workflows/container-image.yml
```diff
@@ -39,202 +37,155 @@ jobs:
...
 
   build-toolbox:
+    strategy:
+      matrix:
+        os: [centos, fedora, opensuse]
+        arch: [amd64]
     needs: build-client
     runs-on: ubuntu-latest
-    env:
-      BUILDAH_FORMAT: oci
```
anoopcs9, line 139: For consistency we could remove above environment variable setting from the entire YAML.

anoopcs9 (changes requested):

tests/test-samba-container.sh
synarete: @obnoxxx Following my previous comments, as well as @anoopcs9 latest comment, I realized that the problem lays in wrong indentations this script had prior to this PR. I took the liberty to re-format it using helper functions (on-top of your PR). IMHO, it fixes both indentation problems and improves readability: 

https://github.com/samba-in-kubernetes/samba-container/commit/0b8988668570f1a162c848a5847dfa7fcf4424b9

obnoxxx: > We could remove extra indentation(`tab`) starting from LINENO:25 through LINENO:47.

fixed, thanks!

tests/test-samba-container.sh
```diff
@@ -1,44 +1,71 @@
...
-		# give samba a second to come up
-		sleep 1
+# give samba a second to come up
+sleep 1
 
-		${CONTAINER_CMD} exec "${CONTAINER_ID}" smbclient -U% -L 127.0.0.1
-		rc=$?
+echo "Listing samba shares"
+	${CONTAINER_CMD} exec "${CONTAINER_ID}" smbclient -U% -L 127.0.0.1
+	rc=$?
```
anoopcs9, line 42:
```suggestion
${CONTAINER_CMD} exec "${CONTAINER_ID}" smbclient -U% -L 127.0.0.1
rc=$?
```

anoopcs9 (changes requested):

dmulder:
> It has already uncovered one issue with the image(s): a missin `/etc/nsswitch.conf` file. we work around this with [2d5dacb](https://github.com/samba-in-kubernetes/samba-container/commit/2d5dacb6d403630f9c44e81232621f70f9b1eb1c).

I also encountered that issue, which is the reason for the glibc dependency (it owns/provides /etc/nsswitch.conf, or did anyway). Maybe that has changed in a recent Tumbleweed snapshot.

tests/test-samba-container.sh
obnoxxx: strange - I thought I had fixed those. anyway: committed now. I'm going to squash these changes into an earlier commit

obnoxxx:
@dmulder  wrote:

> > It has already uncovered one issue with the image(s): a missing `/etc/nsswitch.conf` file. we work around this with [2d5dacb](https://github.com/samba-in-kubernetes/samba-container/commit/2d5dacb6d403630f9c44e81232621f70f9b1eb1c).
> 
> I also encountered that issue, which is the reason for the glibc dependency (it owns/provides /etc/nsswitch.conf, or did anyway). Maybe that has changed in a recent Tumbleweed snapshot.

Thanks for looking at this, @dmulder. And since you are aware of the issue and the root cause, do you suggest a different/better fix or workaround?

tests/test-samba-container.sh
obnoxxx: > strange - I thought I had fixed those. anyway: committed now. I'm going to squash these changes into an earlier commit


done, by the way

obnoxxx:
This PR is in a somewhat special situation because of the following chicken-and-egg problem:

* The PR  adds new CI checks by expanding pe-existing checks with the newly introduced (OS)x(Arch) matrix into multiple checks, essentially removing the original non-matrix checks
* github branch protection settings configure an explicit list of required checks.  For the merge button to become click-able, not only does the PR need two approving reviews but additionally, all required checks must have succeeded. The original, currently configured required checks can not pass since they are removed. but the newly added checks should only be added to the required list after the PR has been merged. otherwise, this would block any other PRs.
* similarly, in order for mergify to automatically merge the PR. the mergify config would need to be expanded by adding the new checks to the required list there. it can only be done after the checks definition has been added *this PR) since adding it earlier would result in an invalid mergify config referencing nonexisting checks. this config update *could* be done as part of this PR without breaking anything but for  the sake of cleanliness is breferably don in a separate follow-up PR.


so since automatic merging and merge-ability is difficult to achieve or at least  undesirable, the path forward seems to be to get two approvals and manually merge with adminitrative  override privileges an change branch protection and mergify config after that.

.github/workflows/container-image.yml
anoopcs9: I can still see that the environment variable getting removed.

.github/workflows/container-image.yml
obnoxxx: @anoopcs9 , we had to exclude centos  from som jobs, specifically those building ad-server images, because the jobs failed as we seem to never never have committed the  corresponding containerfile to the master branch. I Agree that if possible, we should add centos, but it will require some extra work to make it for ad-server initially. I am quite sure that I had that in my original branch and PR to support building with different base OS, but it seems to never have landed

anoopcs9: Yes, that's exactly what I commented above. **build-nightly-server** will work for `centos` and **build-nightly-ad-server** won't currently work for `centos`. I've raised #142 to add Containerfile for CentOS to build ad-server image with nightly samba builds.

anoopcs9 (approved):
lgtm, thanks.

synarete (approved):

PR #138, opened 2023-06-19
makefile: prevent passing an unknown --arch argument to docker

phlogistonjohn:
Assert that if the CONTAINER_CMD is docker that the --arch flag will not be passed to the container engine. Only pass an --arch flag (to podman) if BUILD_ARCH is set to some non-empty value.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,6 +115,11 @@ get_imagename=$(firstword $(subst :, ,$1))
 get_pkgsource=$(if $(findstring nightly,$1),nightly,default)
 
 
+arch_flag=$(strip $(if $(filter docker,$(CONTAINER_CMD)),\
+	$(if $(filter-out $(HOST_ARCH),$(BUILD_ARCH)),\
+		$(error Setting BUILD_ARCH != $(HOST_ARCH) not supported on docker)),\
+	$(if $(BUILD_ARCH),--arch $(BUILD_ARCH))))
+
 build: build-server build-nightly-server build-ad-server build-client \
 	build-toolbox
 .PHONY: build
@@ -305,7 +310,7 @@ check-gitlint: $(filter $(ALT_BIN)%,$(GITLINT))
 _img_build: $(DIR)/.common
 	$(BUILD_CMD) \
 		$(BUILD_ARGS) \
-		$(if $(filter-out $(HOST_ARCH),$(BUILD_ARCH)),--arch $(BUILD_ARCH)) \
+		$(call arch_flag) \
 		$(EXTRA_BUILD_ARGS) \
 		--tag $(SHORT_NAME) \
 		--tag $(REPO_NAME) \


Makefile
```diff
@@ -115,6 +115,11 @@ get_imagename=$(firstword $(subst :, ,$1))
 get_pkgsource=$(if $(findstring nightly,$1),nightly,default)
 
 
+arch_flag=$(strip $(if $(filter docker,$(CONTAINER_CMD)),\
+	$(if $(filter-out $(HOST_ARCH),$(BUILD_ARCH)),\
+		$(error Setting BUILD_ARCH != $(HOST_ARCH) not supported on docker)),\
+	$(if $(BUILD_ARCH),--arch $(BUILD_ARCH))))
```
anoopcs9, line 121: How about reversing the order of `$(filter docker,$(CONTAINER_CMD))` and `$(filter-out $(HOST_ARCH),$(BUILD_ARCH))` conditionals? That would help us in adding `--arch` to **podman** only when `$(HOST_ARCH)` and `$(BUILD_ARCH)` differ. This was pretty much the previous logic.

```suggestion
arch_flag=$(strip $(if $(filter-out $(HOST_ARCH),$(BUILD_ARCH)),\
	$(if $(filter docker,$(CONTAINER_CMD)),\
		$(error Setting BUILD_ARCH != $(HOST_ARCH) not supported on docker),\
	        --arch $(BUILD_ARCH))))
```

I know it doesn't harm to add `--arch` always to **podman** but this way we keep it same for both.

anoopcs9 (changes requested):

synarete (approved):
lgtm

Makefile
anoopcs9: Hm.. that's an interesting find.

Looking more closely at [podman-build(1)](https://docs.podman.io/en/latest/markdown/podman-build.1.html#arch-arch) clearly mentions the following:

> Unless overridden, subsequent lookups of the same image in the local storage matches this architecture, regardless of the host.

..and the reasoning is mentioned in the description of https://github.com/containers/podman/pull/15409.

> Podman has a somewhat irritating behavior that if you build with `--arch arm64` on an amd64 host and then rerun the same command without `--arch` then it will not rebuild.

Despite the above behaviour I see that this situation can still occur with current changes. Consider the following sequence with `podman`:

```
make build-server BUILD_ARCH=arm64 --> builds arm64 image with --arch=arm64
make build-server                  --> tries to build without --arch but finds previously built image
```

phlogistonjohn: Yes, I tried to make the behavior match.

anoopcs9: Then in that regard I would say my original suggestion is still relevant. Or did I miss something?

phlogistonjohn: It's possible we're talking past each other. But I don't know where the disconnect is. I'd prefer to leave this as it currently is for this PR unless you see a bug rather than a suboptimality. Is that OK?

anoopcs9: Ok.

anoopcs9 (approved):

PR #139, opened 2023-06-27
images: support a SAMBACC_VERSION_SUFFIX to force a sambacc version

phlogistonjohn:
In the previous release we made sure to "pin" a version of both sambacc and samba. The former was done by an "embedded" build which is something we've moved away from. So this adds back the ability to require a sambacc version in a way more similar to how we previously required a samba version.


synarete (approved):
lgtm

obnoxxx (approved):
lgtm

PR #147, opened 2023-07-18
github: switch to using fqin style tags in the ci yaml

phlogistonjohn:
These tags are designed to not overlap on major build parameters. This is a stepping stone PR, my next goal is to move the package source into the matrix (default or nightly). But I like a small delta for reviews and this tests out the CI changes with this small step first.


anoopcs9 (approved):
Clean change, lgtm.

synarete (approved):
lgtm

PR #148, opened 2023-07-19
github: combine nightly jobs with default jobs using matrix

phlogistonjohn:
At this point the job definitions for build-nightly-* and test-nightly-* were nearly the same as the non-nightly counterparts. Combine the job definitions via a new matrix parameter.


phlogistonjohn:
I updated the matrix and the matching rules in mergify config. However, this will need to have manual changes made to the github ui too, like the last set of changes that added the matrix.

synarete (approved):
lgtm

.github/workflows/container-image.yml
```diff
@@ -147,83 +165,27 @@ jobs:
...
-      - name: Build the nightly server image
-        run: make KIND=server PACKAGE_SOURCE=nightly OS_NAME=${{ matrix.os }} BUILD_ARCH=${{ matrix.arch }} build-image
-      - name: Upload nightly server image
-        uses: ishworkh/container-image-artifact-upload@v1.0.0
-        with:
-          image: "samba-server:${{ env.IMG_TAG }}"
-          container_engine: ${{ env.CONTAINER_CMD }}
-          retention_days: 1
-
-  build-nightly-ad-server:
```
obnoxxx, line 172: shouldn't this be removed due to the introduction of `build-ad-server(nightly, *, *)`

obnoxxx (changes requested):
this looks very good in general, thanks! But I think the definition of the `build-nightly-server` and `build-nightly-ad-server` jobs should be removed to make this PR complete.

.github/workflows/container-image.yml
```diff
@@ -147,83 +165,27 @@ jobs:
       - name: Test the server image
         run: LOCAL_TAG=samba-server:${{ env.IMG_TAG }} tests/test-samba-container.sh
 
-  # Reminder: the nightly-server images consume nightly samba rpm builds
-  # it is not *just* an image that gets built nightly
-  build-nightly-server:
```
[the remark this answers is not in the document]
obnoxxx: > It is. This is a removed section... see the '-' prefix, and github colors it with a shade of red.

gosh, I am sorry! I looked wrongly...

.github/workflows/container-image.yml
obnoxxx: withdrawinbg my change request as it referred to a non-issue

obnoxxx (approved):
LGTM, great improvement, Thanks!

PR #149, opened 2023-07-19
reformat yaml files to remove yamllint warnings

phlogistonjohn:
The title says it all. Tedious but straightforward.  Gets rid of the noise when running `make check-yaml` so you can more easily see if you've introduced an error or warning.
I disagree with yaml lint on certain trailing comments, but I just went with the flow for now.


synarete (approved):
Looks good. As we are already doing those cleanups, I would also recommend adding re-format to github's action yaml:

`yq eval --inplace .github/workflows/container-image.yml`

anoopcs9 (approved):

PR #150, opened 2023-07-21
gitlint: allow commit messages to contain long footnotes

phlogistonjohn:
Configure gitlint to ignore lines that start like `[1]: ` or `[2] ` such that we can have  long text like URLs in the commits without triggering other gitlint body     rules.

PR contains two test commits to demonstrate the change that must be removed prior to merging.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -30,6 +30,9 @@ ignore-merge-commits=true
 # Enable debug mode (prints more output). Disabled by default.
 # debug=true
 
+# Enable search regex and remove warning message.
+regex-style-search=true
+
 # Enable community contributed rules
 # See http://jorisroovers.github.io/gitlint/contrib_rules for details
 # contrib=contrib-body-requires-signed-off-by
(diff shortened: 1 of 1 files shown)


spuiuk (approved):
ACK - I am all for bending the rules for better readability.

obnoxxx (changes requested):
lgtm. But I think the two  "deleteme" commits should be removed before merging

.gitlint
```diff
@@ -106,11 +109,16 @@ regex=^.*
...
-# [ignore-body-lines]
+[ignore-body-lines]
 # Ignore certain lines in a commit body that match a regex.
 # E.g. Ignore all lines that start with 'Co-Authored-By'
 # regex=^Co-Authored-By
 
+# ignore lines that are "footnotes", that start like `[1]: ` or `[2]: ` and so on
+# this will make it easy to put long urls in commit messages without
+# triggering gitlint body rules
+regex=^\[[0-9]+\]: +
```
anoopcs9, line 120: I would suggest to leave colon(`:`) out as a requirement in the regex pattern for footnotes.

phlogistonjohn: OK, it is now optional.

phlogistonjohn:
This PR also enables the extremely useful feature of being able to put ascsii art in the message witout tickling gitlint. Just prefix each line with `[0]` or something and go to town, like so:
```
stuff: useful things

This is an important commit message.

[0]  ___________________
[0] < this commit rocks >
[0]  ------------------- 
[0]    \
[0]     \
[0]         .--.
[0]        |o_o |
[0]        |:_/ |
[0]       //   \ \
[0]      (|     | )
[0]     /'\_   _/`\
[0]     \___)=(___/
[0]    
[0]                           PS. Please don't actually do this [1].
[0]
[0]
[1]                                      Unless your ascii art is actually a diagram or something
```

anoopcs9 (approved):
lgtm.

spuiuk (approved):
ACK

obnoxxx (approved):
LGTM, thanks!

PR #158, opened 2023-11-02
xxx: do not merge

phlogistonjohn:
Just to check if there are any problems beyond the python3-Markdown dependency. Hopefully we will get a fixed version of opensuse's samba package instead.


phlogistonjohn:
Not needed

PR #164, opened 2024-01-10
hack: some tweaks for the build-image script

phlogistonjohn:
Avoid passing the --arch option to podman for building the host architecture.

Add an option to skip adding tags meant primarily for backwards compatibility with the earlier makefile based builds.

### DIFF
--- a/hack/build-image
+++ b/hack/build-image
@@ -48,6 +48,12 @@ import sys
 
 logger = logging.getLogger("build-image")
 
+# Set FORCE_ARCH_FLAG if you want to test passing the --arch flag to podman all
+# the time. This was the previous behavior but we found it to have some issues.
+# When this is false the --arch flag is passed to podman ONLY when the target
+# arch != the host system arch.
+FORCE_ARCH_FLAG = False
+
 # IMAGE_KINDS - map aliases/names to canonical names for the kinds
 # of images we can build
 IMG_SERVER = "samba-server"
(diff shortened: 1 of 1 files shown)


hack/build-image
```diff
@@ -196,7 +198,13 @@ def container_build(cli, target):
     if "docker" in args[0]:
         if target.arch != host_arch():
             raise RuntimeError("Docker does not support --arch")
-    else:
+    elif target.arch != host_arch() or FORCE_ARCH_FLAG:
+        # We've noticed a few small quirks when using podman with the --arch
+        # option. The main issue is that building the client image works
```
anoopcs9, line 203: Were you able to find any links to existing issues that can be recorded here?

hack/build-image
```diff
@@ -48,6 +48,8 @@ import sys
 
 logger = logging.getLogger("build-image")
 
+FORCE_ARCH_FLAG = False
```
anoopcs9, line 51: What's the real intention here if `FORCE_ARCH_FLAG` is unchanged? Is that some kind of placeholder in case the issue with `--arch` and podman is resolved?

synarete (approved):
Looks fine. 

However: top-level doc-string still implies using explicit `--arch` option. Better to modify the doc string as well to avoid confusion in the future.

hack/build-image
phlogistonjohn: No, I just kept hitting this while working on the releases. I probably should look up if there's a podman issue for this and or report it. Regardless, the issue exists.

phlogistonjohn:
> However: top-level doc-string still implies using explicit `--arch` option. Better to modify the doc string as well to avoid confusion in the future.

I'm not sure what you mean. if you mean the doc-comment at lines 2-36 those are the --arch flags that are passed to this script to set the target arch. With the new change if you don't supply `--arch` or the value of `--arch` option to build-image is the same as the host's architecture we'll skip passing an `--arch` to podman.

synarete:
> > However: top-level doc-string still implies using explicit `--arch` option. Better to modify the doc string as well to avoid confusion in the future.
> 
> I'm not sure what you mean. if you mean the doc-comment at lines 2-36 those are the --arch flags that are passed to this script to set the target arch. With the new change if you don't supply `--arch` or the value of `--arch` option to build-image is the same as the host's architecture we'll skip passing an `--arch` to podman.

Fair enough.

hack/build-image
anoopcs9: In that case should we add a comment here? Otherwise first impression would be similar to mine where it seems kind of useless.

phlogistonjohn: Done.

phlogistonjohn:
@anoopcs9 @synarete PTA2L

anoopcs9 (approved):
lgtm.

synarete (approved):
lgtm

PR #165, opened 2024-01-12
Add ceph support for centos based images

phlogistonjohn:
Ensure ceph support exists in the centos based images. Ceph support (for the samba vfs module and sambacc) was already present in the Fedora based images due to (a) fedora packages cephs in the main repos (b) the samba and sambacc RPMs having correct dependencies, as long as the proper samba  vfs modules are selected. These changes will enable ceph libraries to be installed on the samba server image when based on centos and works around some issues in the sambacc dependencies as well.

Add a new package source called "devbuilds" that is valid for centos based images. It enables both the samba nightly builds repo (like the existing "nightly" package source) as well as dev builds from ceph's main branch.

In order to do this cleanly the script and images are refactored slightly to re-organize how packages are selected for install.

Finally, enable ceph builds from the reef branch on centos nightly - note that there's a bug in the current version of reef that prevents the client from talking to newer mds servers but that will eventually get resolved (Hopefully).

### DIFF
--- a/images/server/install-packages.sh
+++ b/images/server/install-packages.sh
@@ -13,6 +13,7 @@ get_custom_repo() {
 install_packages_from="$1"
 samba_version_suffix="$2"
 install_custom_repo="$3"
+package_selection="$4"
 
 # shellcheck disable=SC1091
 OS_BASE="$(. /etc/os-release && echo "${ID}")"
@@ -20,6 +21,7 @@ OS_BASE="$(. /etc/os-release && echo "${ID}")"
 case "${install_packages_from}" in
     samba-nightly)
         get_custom_repo "https://artifacts.ci.centos.org/samba/pkgs/master/${OS_BASE}/samba-nightly-master.repo"
+        package_selection=${package_selection:-nightly}
     ;;
     custom-repo)
         get_custom_repo "${install_custom_repo}"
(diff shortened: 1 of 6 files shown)


images/server/install-packages.sh
```diff
@@ -2,26 +2,67 @@
 
 set -ex
 
+
+need_curl() {
+    if command -v curl >/dev/null ; then
+        return 0
+    fi
+    dnf install --setopt=install_weak_deps=False -y /usr/bin/curl
```
synarete, line 10: Why do you need the `return 0`? maybe just have this one-liner:
```
command -v curl > /dev/null || dnf install -y --setopt=install_weak_deps=False curl
```

images/server/install-packages.sh
```diff
@@ -2,26 +2,67 @@
...
+import json
+r = json.load(open(json_file))
+url = r[0]["url"]
+ref = r[0]["ref"]
+with open(dest, "w") as out:
+    print(f"[ceph-{ref}]", file=out)
+    print(f"name=Ceph Development Build ({ref})", file=out)
+    print(f"baseurl={url}/x86_64", file=out)
+    print("enabled=1", file=out)
+    print("gpgcheck=0", file=out)
```
synarete, line 38: For me, when I get to the point where I need to embed Python code with shell script, its a sign that I should consider re-writing the entire script in python. It would probably be more maintainable in the long run. (just a though, not a specific action for this PR).

phlogistonjohn: Yeah.... it certainly crossed my mind, but the python here is pretty isolated to just extracting the needed data from json. In theory, I could have used jq but that's not already part of the image and python is.

images/server/install-packages.sh
phlogistonjohn: Sure. I will keep it in mind as an optimization if I need to touch the script for any other reason.

synarete: Fair enough.

obnoxxx (commented):
This looks nice and clean from a first superficial review.

I am inclined to approve but should look at the code changes in more detail.

What seems to be a little hidden  among all the preparatory commits is the fact that the main achievement of the PR according to the tiele, i. e. adding support for ceph, is essentially achieved by  affing the samba-vfs-ceph package and dependencies from ceph itself. IMHO, this could be made clearer in the PR description and corresponding commit message. At this stage, this is not a formal change request but merely a comment 😄

phlogistonjohn:
I added a new intro paragraph to the description. I hope that helps.

obnoxxx (approved):
Thanks for updating the description, @phlogistonjohn ! This is much clearer now. LGTM

phlogistonjohn:
@synarete would you mind approving if you're OK with my earlier responses?

synarete (approved):

PR #166, opened 2024-01-23
workflows: restructure the push job

phlogistonjohn:
The following changes update the build-image script so that it's easier to update the tags for existing images and helps control what tags are pushed when the push command is invoked.

Then we restructure the push job so that it avoids rebuilding the images we just built and test and instead reimports them from the artifacts and then uses the new retag command to fix up the tags and distro base followed by pushing them as both unqualified (`latest`, `nightly`) tags and fully-qualified tags.

Finally, we enable pushing the `samba-server:nightly-centos-amd64` and `samba-server:devbuilds-centos-amd64` to quay.io. The latter will be used for testing the ongoing SMB on Ceph work occurring in the Ceph project.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -226,7 +226,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: log in to quay.io
-        run: docker login -u "${{ secrets.QUAY_USER }}" -p "${{ secrets.QUAY_PASS }}" quay.io
+        run: ${CONTAINER_CMD} login -u "${{ secrets.QUAY_USER }}" -p "${{ secrets.QUAY_PASS }}" quay.io
       - name: push server image
         run: make push-server
       - name: push ad-server image
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -247,6 +247,16 @@ jobs:
         with:
           image: "samba-server:nightly-fedora-amd64"
           container_engine: ${{ env.CONTAINER_CMD }}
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
FWIW I tested this on my fork of samba-container: you can see some of the results (modulo hacks I did in order to point it at unofficial repos) here: https://github.com/phlogistonjohn/samba-container/actions

synarete (approved):
Luckily, build-image is written in Python ;)

lgtm.

.github/workflows/container-image.yml
```diff
@@ -222,20 +228,87 @@ jobs:
...
+      - name: Fetch ad-server default-fedora-amd64
+        uses: ishworkh/container-image-artifact-download@v1.0.0
+        with:
+          image: "samba-ad-server:default-fedora-amd64"
+          container_engine: ${{ env.CONTAINER_CMD }}
+      - name: Fetch ad-server nightly-fedora-amd64
+        uses: ishworkh/container-image-artifact-download@v1.0.0
+        with:
+          image: "samba-ad-server:nightly-fedora-amd64"
+          container_engine: ${{ env.CONTAINER_CMD }}
```
anoopcs9, line 270: Aren't we interested in `samba-ad-server:nightly-centos-amd64`?

anoopcs9: > I can add it later but I don't need it for ceph work ATM.

👍🏻 

> The AD server image is a bit more self contained (no pun intended) and works OK when paired w/ a centos based server image.

I'm not sure I understand this part but its OK for now.

PR #167, opened 2024-02-06
images/ad-server: Send debug output to standard output

anoopcs9:
_/usr/sbin/samba_ also comes with a similar option as others to print logs to standard output. Make use of `SAMBA_SPECIFICS` argument to specify required value to finally run with `--debug-stdout` option.

depends on https://github.com/samba-in-kubernetes/sambacc/pull/106

### DIFF
--- a/images/ad-server/Containerfile.centos
+++ b/images/ad-server/Containerfile.centos
@@ -2,7 +2,7 @@ FROM quay.io/centos/centos:stream9
 ARG INSTALL_PACKAGES_FROM=default
 ARG SAMBA_VERSION_SUFFIX=""
 ARG SAMBACC_VERSION_SUFFIX=""
-ARG SAMBA_SPECIFICS=
+ARG SAMBA_SPECIFICS=daemon_cli_debug_output
 ARG INSTALL_CUSTOM_REPO=
 
 MAINTAINER John Mulligan <jmulligan@redhat.com>
--- a/images/ad-server/Containerfile.fedora
+++ b/images/ad-server/Containerfile.fedora
@@ -2,7 +2,7 @@ FROM registry.fedoraproject.org/fedora:39
 ARG INSTALL_PACKAGES_FROM=default
 ARG SAMBA_VERSION_SUFFIX=""
 ARG SAMBACC_VERSION_SUFFIX=""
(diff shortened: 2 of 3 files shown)


images/ad-server/Containerfile.centos
```diff
@@ -2,7 +2,7 @@ FROM quay.io/centos/centos:stream9
 ARG INSTALL_PACKAGES_FROM=default
 ARG SAMBA_VERSION_SUFFIX=""
 ARG SAMBACC_VERSION_SUFFIX=""
-ARG SAMBA_SPECIFICS=
+ARG SAMBA_SPECIFICS=daemon_cli_debug_output
```
phlogistonjohn, line 5: I think that the centos container file didn't have this value for SAMBA_SPECIFICS because when we started out with centos builds in samba-container, the samba version on centos didn't have daemon_cli_debug_output (but Fedora versions did).

This is just a double check - and I'll trust you - that you've checked that the default packages for centos (non-nightly) support those options now?

anoopcs9: > I think that the centos container file didn't have this value for SAMBA_SPECIFICS because when we started out with centos builds in samba-container, the samba version on centos didn't have daemon_cli_debug_output (but Fedora versions did).

It might also have been a factor that we initially had CentOS Stream 8 and not CentOS Stream 9(with latest samba versions) as base image.

> This is just a double check - and I'll trust you - that you've checked that the default packages for centos (non-nightly) support those options now?

At present both Stream 8 and Stream 9 has higher samba version than minimum required in this case.
```
# dnf list samba
Available Packages
samba.x86_64                                                                                4.19.4-3.el8                                                                                 baseos

Available Packages
samba.x86_64                                                                               4.19.3-100.el9                                                                                baseos
```

anoopcs9: Re-ran the CI to pick latest sambacc which completed successfully.

One more realization: We don't build samba-ad-server with CentOS default packages due to unavailability of Samba DC RPMs 😝

phlogistonjohn (approved):
lgtm

synarete (approved):
lgtm

PR #168, opened 2024-03-04
workflows: allow manually running the ci job

phlogistonjohn:
Add a workflow_dispatch event trigger so that the team (specifically, me when I am being impatient) can trigger a manual CI run and eventually image push to quay.io


synarete (approved):
lgtm

There are a lot of configuration setting we may enable with `workflow_dispatch`. I assume we may leave it for future PR.

obnoxxx (approved):
I wasn't aware of this mechanism and had to look it up, but this looks clean. LGTM

PR #174, opened 2024-07-04
images/server: Install ctdb-ceph-mutex for rados helper

anoopcs9:
`ctdb-rados-mutex` subcommand to `samba-container` expects the required rados helper binary to be present for execution. Therefore install the necessary ctdb-ceph-mutex package providing ctdb_mutex_ceph_rados_helper in various image flavours.

### DIFF
--- a/images/server/install-packages.sh
+++ b/images/server/install-packages.sh
@@ -95,11 +95,11 @@ samba_packages=(\
     ctdb)
 case "${package_selection}-${OS_BASE}" in
     *-fedora|allvfs-*)
-        samba_packages+=(samba-vfs-cephfs samba-vfs-glusterfs)
+        samba_packages+=(samba-vfs-cephfs samba-vfs-glusterfs ctdb-ceph-mutex)
     ;;
     nightly-centos|devbuilds-centos|forcedevbuilds-*)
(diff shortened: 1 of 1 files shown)


images/server/install-packages.sh
```diff
@@ -95,11 +95,11 @@ samba_packages=(\
     ctdb)
 case "${package_selection}-${OS_BASE}" in
     *-fedora|allvfs-*)
-        samba_packages+=(samba-vfs-cephfs samba-vfs-glusterfs)
```
phlogistonjohn, line 98: I think we had gluster enabled on fedora in case someone in the community wanted to tinker. If you think that's not worthwhile, that's fine with me as I didn't plan on using it.

anoopcs9: Fine. The retirement plan is for Fedora 42 only. I've decided to keep it for another year(maybe).

phlogistonjohn: "TIL" about retiring glusterfs from fedora. I'm OK with removing it I just didn't want to remove it as a side effect of adding something else unrelated. :-)

phlogistonjohn (approved):
lgtm

PR #175, opened 2024-08-19
workflows: allow push on manually run workflows

phlogistonjohn:
A while back we enabled running workflow manually in the action (with a button on the github ui) via the workflow_dispatch type. However, this action doesn't run the 'push' step because push is only included with a scheduled run or a merged pr. Rework the line so that push is performed for scheduled runs, merged prs, or manual runs. This is useful in case something samba-container depends on is updated and one doesn't feel like waiting all day for a scheduled run.


obnoxxx (approved):
LGTM, thanks!

Issue #155, opened 2023-08-10
Add AARCH64 Support


SeboKnt:
Hi, on my K8s cluster on the x86/Ubuntu nodes it works fine (thanks for that) but it's a multi-arch cluster and I can't migrate the container to my arm64/Alpine nodes. Are there any plans to make the container available for arm architecture?

apexkode:
Hi I would also like support for arm64 arch, however I am struggling to build with the instructions provided, is there a timeline for when arm64 builds will be available on quay.io?

anoopcs9:
> Hi I would also like support for arm64 arch, however I am struggling to build with the instructions provided,

What are the issues you have seen with podman by running `make BUILD_ARCH=arm64 build-server`?

apexkode:
@anoopcs9  I encountered the same issue described here: https://github.com/containers/podman/issues/18483
Which does not seem to be directly related to this project but rather an issue in podman and Fedora (possibly), but did not yet try the work around which was suggested as I did not want to [rm -rf](https://github.com/containers/podman/issues/18483#issuecomment-1536054770) my container directory, I was going to look further into this when I had more time.

phlogistonjohn:
There's no timeline, sorry.

I don't know if there's a good way to work around the podman issue. I understand your reluctance to simply delete your containers directory. You probably thought of this already, but maybe you could find a different spare node to build on?

FWIW I'd be happy to work with any one who can make contributions to the build scripts in order to start building for arm64 as most of the infrastructure should be there. I just personally don't have the time for it myself right now.

mogeko:
Nowadays, the Arm-based hosted runners in GitHub Actions is in Beta. Perhaps in the near future, we will see the samba-container working on the Arm node.

See:

- https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions
- https://resources.github.com/devops/accelerate-your-cicd-with-arm-and-gpu-runners-in-github-actions

PR #180, opened 2024-09-13
build Fedora+OpenSUSE arm64 container images

abachmann:
Extends the GA workflow to build and push Fedora and OpenSUSE arm64 container images. The `build-image` script has been modified to use Docker Buildx for cross-building.

Preparation for: https://github.com/samba-in-kubernetes/samba-container/pull/179
Fixes: https://github.com/samba-in-kubernetes/samba-container/issues/155

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -45,17 +45,17 @@ jobs:
       matrix:
         package_source: [default, nightly]
         os: [centos, fedora, opensuse]
-        arch: [amd64]
+        arch: [amd64, arm64]
         exclude:
           # there are no nightly packages for opensuse
           - package_source: nightly
             os: opensuse
+          - os: centos
+            arch: arm64
         include:
           - package_source: devbuilds
             os: centos
             arch: amd64
-          - os: fedora
-            arch: arm64
     runs-on: ubuntu-latest
     env:
       BUILDAH_FORMAT: oci
@@ -76,7 +76,7 @@ jobs:
       matrix:
         package_source: [default, nightly]
         os: [centos, fedora, opensuse]
-        arch: [amd64]
+        arch: [amd64, arm64]
         exclude:
           # there are no nightly packages for opensuse
           - package_source: nightly
@@ -84,8 +84,7 @@ jobs:
           # the distro packages for centos do not include an ad-dc
           - package_source: default
             os: centos
-        include:
-          - os: fedora
+          - os: centos
             arch: arm64
     runs-on: ubuntu-latest
     env:
@@ -106,9 +105,9 @@ jobs:
     strategy:
       matrix:
         os: [centos, fedora, opensuse]
-        arch: [amd64]
-        include:
-          - os: fedora
+        arch: [amd64, arm64]
+        exclude:
+          - os: centos
             arch: arm64
     runs-on: ubuntu-latest
     env:
@@ -263,6 +262,11 @@ jobs:
         with:
           image: "samba-server:default-fedora-arm64"
           container_engine: ${{ env.CONTAINER_CMD }}
+      - name: Fetch server default-opensuse-arm64
+        uses: ishworkh/container-image-artifact-download@v2.0.0
+        with:
+          image: "samba-server:default-opensuse-arm64"
+          container_engine: ${{ env.CONTAINER_CMD }}
       - name: Fetch server nightly-fedora-amd64
         uses: ishworkh/container-image-artifact-download@v2.0.0
         with:
@@ -294,6 +298,11 @@ jobs:
         with:
           image: "samba-ad-server:default-fedora-arm64"
           container_engine: ${{ env.CONTAINER_CMD }}
+      - name: Fetch ad-server default-opensuse-arm64
+        uses: ishworkh/container-image-artifact-download@v2.0.0
+        with:
+          image: "samba-ad-server:default-opensuse-arm64"
+          container_engine: ${{ env.CONTAINER_CMD }}
       - name: Fetch ad-server nightly-fedora-amd64
         uses: ishworkh/container-image-artifact-download@v2.0.0
         with:
@@ -315,6 +324,11 @@ jobs:
         with:
           image: "samba-client:default-fedora-arm64"
           container_engine: ${{ env.CONTAINER_CMD }}
+      - name: Fetch client default-opensuse-arm64
+        uses: ishworkh/container-image-artifact-download@v2.0.0
+        with:
+          image: "samba-client:default-opensuse-arm64"
+          container_engine: ${{ env.CONTAINER_CMD }}
       # (toolbox images)
       - name: Fetch toolbox default-fedora-amd64
         uses: ishworkh/container-image-artifact-download@v2.0.0
@@ -331,16 +345,19 @@ jobs:
           --no-distro-qualified
           -i samba-server:default-fedora-amd64
           -i samba-server:default-fedora-arm64
+          -i samba-server:default-opensuse-arm64
           -i samba-server:nightly-fedora-amd64
           -i samba-server:nightly-fedora-arm64
           -i samba-server:nightly-centos-amd64
           -i samba-server:devbuilds-centos-amd64
           -i samba-ad-server:default-fedora-amd64
           -i samba-ad-server:default-fedora-arm64
+          -i samba-ad-server:default-opensuse-arm64
           -i samba-ad-server:nightly-fedora-amd64
           -i samba-ad-server:nightly-fedora-arm64
           -i samba-client:default-fedora-amd64
           -i samba-client:default-fedora-arm64
+          -i samba-client:default-opensuse-arm64
           -i samba-toolbox:default-fedora-amd64
       - name: Push images
         run: >
@@ -352,12 +369,15 @@ jobs:
           --push-selected-tags=mixed
           -i ${REPO_BASE}/samba-server:default-fedora-amd64
           -i ${REPO_BASE}/samba-server:default-fedora-arm64
+          -i ${REPO_BASE}/samba-server:default-opensuse-arm64
           -i ${REPO_BASE}/samba-server:nightly-fedora-amd64
           -i ${REPO_BASE}/samba-server:nightly-centos-amd64
           -i ${REPO_BASE}/samba-server:devbuilds-centos-amd64
           -i ${REPO_BASE}/samba-ad-server:default-fedora-amd64
           -i ${REPO_BASE}/samba-ad-server:default-fedora-arm64
+          -i ${REPO_BASE}/samba-ad-server:default-opensuse-arm64
           -i ${REPO_BASE}/samba-ad-server:nightly-fedora-amd64
           -i ${REPO_BASE}/samba-client:default-fedora-amd64
           -i ${REPO_BASE}/samba-client:default-fedora-arm64
+          -i ${REPO_BASE}/samba-client:default-opensuse-arm64
           -i ${REPO_BASE}/samba-toolbox:default-fedora-amd64
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -54,6 +54,8 @@ jobs:
           - package_source: devbuilds
             os: centos
             arch: amd64
+          - os: fedora
+            arch: arm64
     runs-on: ubuntu-latest
     env:
       BUILDAH_FORMAT: oci
@@ -82,6 +84,9 @@ jobs:
           # the distro packages for centos do not include an ad-dc
           - package_source: default
             os: centos
+        include:
+          - os: fedora
+            arch: arm64
     runs-on: ubuntu-latest
     env:
       BUILDAH_FORMAT: oci
@@ -102,6 +107,9 @@ jobs:
       matrix:
         os: [centos, fedora, opensuse]
         arch: [amd64]
+        include:
+          - os: fedora
+            arch: arm64
     runs-on: ubuntu-latest
     env:
       BUILDAH_FORMAT: oci
@@ -250,10 +258,20 @@ jobs:
         with:
           image: "samba-server:default-fedora-amd64"
           container_engine: ${{ env.CONTAINER_CMD }}
+      - name: Fetch server default-fedora-arm64
+        uses: ishworkh/container-image-artifact-download@v2.0.0
+        with:
+          image: "samba-server:default-fedora-arm64"
+          container_engine: ${{ env.CONTAINER_CMD }}
       - name: Fetch server nightly-fedora-amd64
         uses: ishworkh/container-image-artifact-download@v2.0.0
         with:
           image: "samba-server:nightly-fedora-amd64"
+          container_engine: ${{ env.CONTAINER_CMD }}
+      - name: Fetch server nightly-fedora-arm64
+        uses: ishworkh/container-image-artifact-download@v2.0.0
+        with:
+          image: "samba-server:nightly-fedora-arm64"
           container_engine: ${{ env.CONTAINER_CMD }}
       - name: Fetch server nightly-centos-amd64
         uses: ishworkh/container-image-artifact-download@v2.0.0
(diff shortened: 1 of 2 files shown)


abachmann:
@phlogistonjohn do you have time to look into this topic? It would be nice to see if the pipeline is running with my changes.

phlogistonjohn:
I am currently traveling and dont have a lot of time too look, but I will allow the CI jobs to run!

obnoxxx (changes requested):
Thanks a lot for your contribution!

This looks like a good addition.

Siome CI checks hafe failed, please fix. 
Especially the check-commits test. (no commit message body provided).

abachmann:
@obnoxxx Thanks for the review. I have added a message body to all commits.

I found a potential reason for the pipeline failure: https://github.com/samba-in-kubernetes/samba-container/pull/180/commits/7795ca5783acbc04d9c0ddeeb88c4126c1856c7b

If it's okay to remove the `arch` parameter, I can squash the following commits:
- https://github.com/samba-in-kubernetes/samba-container/pull/180/commits/bd8e932a1ef052a37e171fba645674cab80e7a04
- https://github.com/samba-in-kubernetes/samba-container/pull/180/commits/b78b782be1e7394d49cfee3d10fc16e0fefbf9ae

abachmann:
Things are getting a bit more interesting now.
`49.85   - Status code: 404 for https://mirrors.centos.org/metalink?repo=centos-resilientstorage-9-stream&arch=aarch64&protocol=https,http (IP: 8.43.85.73)`

The `centos-resilientstorage-9-stream` repository is only available for the following architectures:
- ppc64le
- s390x
- x86_64

It’s also possible to set the arch parameter to source. This will increase build time, but we could add a check: if the target architecture is not supported, then fall back to source. This way, the amd64 builds won’t be affected.

I checked the list of available repositories and their architecture support. A lot of them support aarch64, but not the resilientstorage repositories. So, if we start building the packages from source, we might run into issues.

Before I proceed, I’d like to get your thoughts on this. Maybe you're already aware of the issue and have some solutions in mind.

obnoxxx (changes requested):
Thanks for adding commit message bodies, @abachmann !

If you want to squash the two commits that change the workflows, go ahead. I don't have a strong opinion on that part as long as the change that modifies the script remains a separate commit.

I didn't quite understand yet what makes it necessary to remove the explicit invocation of the arch from the workflows. I do prefer we keep testing arm builds in the ci. 


some ci checks are still failing. please address.

obnoxxx:
@abachmann, any updates here?

We'd really like to take the changes, but please address the requests.

abachmann:
@obnoxxx, , sorry for the late response. I was busy with other things.

> I didn't quite understand yet what makes it necessary to remove the explicit invocation of the arch from the workflows. I do prefer we keep testing arm builds in the ci.

Yeah, I initially thought the explicit invocation of the arch was causing the pipeline issue, but it turned out that wasn’t the case. So, I reverted the commit.

As I wrote in my last comment:
> Things are getting a bit more interesting now.
> 49.85   - Status code: 404 for https://mirrors.centos.org/metalink?repo=centos-resilientstorage-9-> stream&arch=aarch64&protocol=https,http (IP: 8.43.85.73)
>
> The centos-resilientstorage-9-stream repository is only available for the following architectures:
>
> - ppc64le
> - s390x
> - x86_64

Addressing the pipeline issue isn't so easy. I'll try to find some time over the weekend to fix it.

obnoxxx:
@abachmann  wrote:
> @obnoxxx, sorry for the late response. I was busy with other things.

sure. Thanks for following up! And sorry for nagging so badly! 

> > I didn't quite understand yet what makes it necessary to remove the explicit invocation of the arch from the workflows. I do prefer we keep testing arm builds in the ci.
> 
> Yeah, I initially thought the explicit invocation of the arch was causing the pipeline issue, but it turned out that wasn’t the case. So, I reverted the commit.

Thanks for explaining. This makes more sense now.

> 
> As I wrote in my last comment:
> 
> > Things are getting a bit more interesting now.
> > 49.85   - Status code: 404 for https://mirrors.centos.org/metalink?repo=centos-resilientstorage-9-stream&arch=aarch64&protocol=https,http (IP: 8.43.85.73)

right. The centos repo mirrors do sometimes have issues with availability.

> > The centos-resilientstorage-9-stream repository is only available for the following architectures:
> > 
> > * ppc64le
> > * s390x
> > * x86_64

Interesting. 


> 
> Addressing the pipeline issue isn't so easy. I'll try to find some time over the weekend to fix it.

Thanks! Looking forward to seeing a solution.

abachmann:
@obnoxxx
> sure. Thanks for following up! And sorry for nagging so badly!

This topic was completely out of mind, so nagging was actually a good idea ;)

Over the weekend, I had some time to look more deeply into the issue where the `resilientstorage` repository was not found during the image build. As I mentioned before, no `aarch64` packages are provided in the official repositories. So, I tried to build the packages from source. To do this, I modified the [install-packeges.sh](https://github.com/samba-in-kubernetes/samba-container/blob/master/images/server/install-packages.sh#L142) file as follows:
```diff
-     dnf_cmd+=(--enablerepo=crb --enablerepo=resilientstorage)
+     dnf_cmd+=(--enablerepo=crb --enablerepo=resilientstorage-source)
```
This resolved the 404 error, but I ran into another issue afterward: the `ctdb` package, which is provided by `resilientstorage` repository, was not installed. So, building the packages from source didn’t work as expected. Unfortunately, I wasn’t able to find another repository that provides the necessary package.

If you know a repository which provides the missing package, please let me know. Otherwise, I would suggest to skip the `centos aarch64` images for  for now, because all other OSes seem to work. I'll double-check it this week.

phlogistonjohn:
I've had very little time to follow up with you on this but I did want to pop in and say that the idea to start with arm64 on the fedora based images alone would be fine - in fact, highly recommended.

FWIW, while I'm thinking about this topic, I would envision that the github actions build matrix have a task to build the arm64 images as an explicit include, for only fedora with default package source, similar to how we extend the build matrix for the devbuilds image:
https://github.com/samba-in-kubernetes/samba-container/blob/a5244a3568ad70da0768f11cebeb5337ebe043d7/.github/workflows/container-image.yml#L53-L56

One reason is that is is clean (IMO) and another is that I'm 90% sure the nightly builds don't have arm64 versions either. We'd need to work with @anoopcs9 and make changes in the https://github.com/samba-in-kubernetes/samba-build repo before we could consume those nightly samba builds in the images.

nerijus:
> I wasn’t able to find another repository that provides the necessary `arch64` package.

Both AlmaLinux and Rocky Linux have `ctdb` aarch64 packages - https://pkgs.org/search/?q=ctdb

obnoxxx (changes requested):
I agree with @phlogistonjohn in that what is still missing here is making the arch one dimension of the workflow matrix 
here: https://github.com/samba-in-kubernetes/samba-container/blob/a5244a3568ad70da0768f11cebeb5337ebe043d7/.github/workflows/container-image.yml#L53-L56

Additionally, we can limit to Fedora for now and instead of waiting for arm packages from the distos, we should probably add arm builds to our own here: https://github.com/samba-in-kubernetes/samba-build

Finally, do we need to use docker for the container build, or can we use podman instead?

.github/workflows/container-image.yml
```diff
@@ -250,6 +261,11 @@ jobs:
         with:
           image: "samba-server:default-fedora-amd64"
           container_engine: ${{ env.CONTAINER_CMD }}
+      - name: Fetch server default-fedora-arm64
+        uses: ishworkh/container-image-artifact-download@v2.0.0
+        with:
+          image: "samba-server:default-fedora-arm64"
+          container_engine: ${{ env.CONTAINER_CMD }}
```
phlogistonjohn, line 268: I'm going to recommend skipping the tests for now. Let's just get the builds on emulated arm64 in this initial PR.

hack/build-image
```diff
@@ -125,6 +125,7 @@ QUAL_FQIN = 'fqin'
 
 
 _DISCOVERED_CONTAINER_ENGINES = []
+_BUILDX_BUILDER_NAME = "samba-in-kubernetes"
```
phlogistonjohn, line 128: Could you explain what this does? I'm not too familiar with buildx.

hack/build-image
```diff
@@ -219,6 +231,11 @@ def container_build(cli, target):
         # --arch is not provided. So if the target arch and the host_arch
         # are the same, skip passing the extra argument.
         args.append(f"--arch={target.arch}")
+
+    pkgs_from = PACKAGES_FROM[target.pkg_source]
+    if pkgs_from:
+        args.append(f"--build-arg=INSTALL_PACKAGES_FROM={pkgs_from}")
+
```
phlogistonjohn, line 238: I'm not clear on what this addition is doing. Can you elaborate?

hack/build-image
abachmann: When installing buildx, a default builder is created that can only build images for the host platform, which in my case is amd64.

The `docker buildx ls` command can be used to list all available builders:

```sh
NAME/NODE               DRIVER/ENDPOINT             STATUS  BUILDKIT             PLATFORMS
default *               docker                                                   
  default               default                     running v0.11.7+d3e6c1360f6e linux/amd64, linux/amd64/v2, linux/amd64/v3, linux/386
```

To build images for platforms other than the host, a new builder must be created that supports other architectures. Each builder requires a unique name, so I used the GitHub group name as the builder name, and this variable holds its value.

After creating the new builder, samba-in-kubernetes, the output of `docker buildx ls` looks like this:

```sh
NAME/NODE               DRIVER/ENDPOINT             STATUS  BUILDKIT             PLATFORMS
samba-in-kubernetes     docker-container                                         
  samba-in-kubernetes0  unix:///var/run/docker.sock running v0.16.0              linux/arm64*, linux/amd64*, linux/arm/v7*
```

hack/build-image
abachmann: I did not add those lines; I only moved this block down a little bit.

abachmann:
> Finally, do we need to use docker for the container build, or can we use podman instead?

@obnoxxx, I believe building the images with podman should work as well, but I have not tested it.

Finally, the pipeline built the following Fedora arm64 images successfully:
- server
- ad-server
- client

@phlogistonjohn, what about the toolbox image? This image is only used for testing, so we can skip it, right? If we need to build it, then we have an issue because the [Dockerfile ](https://github.com/samba-in-kubernetes/samba-container/blob/master/images/toolbox/Containerfile.fedora) is using the latest tag, which points to the Fedora amd64 image.

hack/build-image
phlogistonjohn: Yes, I missed that, thank you.

hack/build-image
phlogistonjohn: I think I understand the buildx workflow a bit better now, thanks!

hack/build-image
```diff
@@ -203,14 +204,25 @@ def container_engine(cli):
...
+                "build",
+                f"--builder={_BUILDX_BUILDER_NAME}",
+                f"--platform=linux/{target.arch}",
+                "--load"
+            ]
+            # Docker's default builder only supports the host architecture.
+            # Therefore, we need to create a new builder to support other
+            # architectures. Errors are suppressed to prevent issues when
+            # the builder is already available - this can be improved later.
+            run(cli, [eng, "buildx", "create", f"--name={_BUILDX_BUILDER_NAME}"], check=False)
```
phlogistonjohn, line 225: I think it would be a bit clearer if we moved this line and the corresponding comment above the assignment to `args =` (currently on line 213). That way a reader immediately sees the command setting up the new builder

phlogistonjohn: Do we need to remove/clean up this builder? I won't require it for the PR but for people using the script outside the CI environment (and we exist) I wouldn't not like to leave stuff behind. Similarly, is there an issue if this command is executed >1 time on the same host?

phlogistonjohn:
> @phlogistonjohn, what about the toolbox image? This image is only used for testing, so we can skip it, right? If we need to build it, then we have an issue because the [Dockerfile ](https://github.com/samba-in-kubernetes/samba-container/blob/master/images/toolbox/Containerfile.fedora)is using the latest tag, which points to the Fedora amd64 image.

I'm OK with skipping it for now. I'm in favor of moving forward in baby steps for this work :-)

obnoxxx:
I created a request to add arm builds to samba-build: https://github.com/samba-in-kubernetes/samba-build/issues/63

Once we have that, using those RPMs for Arm containers should not be a problem

hack/build-image
abachmann: I added two new commits to clean up the container_build method a little bit and to address the open topics. 

When starting a new container build, the builder, which might exist due to a failure, is removed. This way, we can ensure that we always start with a fresh builder and have no leftovers from a previous build. Afterwards, a new clean builder is created. After the build process is complete, the created builder will be removed.

hack/build-image
abachmann: I removed the global variable and used `target.flat_name()` as the builder name instead.

obnoxxx:
> I created a request to add arm builds to samba-build: [samba-in-kubernetes/samba-build#63](https://github.com/samba-in-kubernetes/samba-build/issues/63)
> 
> Once we have that, using those RPMs for Arm containers should not be a problem

@abachmann , FYI:

RPM builds for ARM have now been implemented in the samba-build repo and er have first nightly builds from master and release 4.21.


https://artifacts.ci.centos.org/samba/pkgs/master/fedora/samba-nightly-master.repo 
https://artifacts.ci.centos.org/samba/pkgs/4.21/fedora/samba-nightly-4.21.repo

https://artifacts.ci.centos.org/samba/pkgs/master/centos/samba-nightly-master.repo
 https://artifacts.ci.centos.org/samba/pkgs/4.21/centos/samba-nightly-4.21.repo


This samba-container project already has mechanisms to use those custom yum repos.

phlogistonjohn:
Because of the work @anoopcs9 did on those builds I learned that centos ci has real arm builders - I was not aware of this before. My suggestion is to drive this PR to completion but once it is done we should rethink how our "nightly" images are built. It may make way more sense to switch to the centos ci infrastructure. We can discuss the topic in more detail later but I don't think building up everything using github actions is ideal path forward when we have access to actual arm builders.

abachmann:
@obnoxxx, thanks for the info. I’ll try to find some time over the weekend to take a closer look.

abachmann:
@phlogistonjohn, I addressed the review concerns in a separate commit to ease the review process. If you're okay with these changes, I'd like to combine the commit [hack: use buildx to cross-build images with docker](https://github.com/samba-in-kubernetes/samba-container/pull/180/commits/028a2494475db1d327197e588552af101a61bea0) and [hack: rework container_build function](https://github.com/samba-in-kubernetes/samba-container/pull/180/commits/4ff545cdc66fc124908ea589a668d82b6f6c9410) into a single commit, as I don't see any benefit to keeping them separate.  I also want to ensure we're all on the same page and fine with the current changes before I start working on the CentOS topic.

abachmann:
@phlogistonjohn, @obnoxxx, I have a question: what about OpenSUSE images? I see that images are built for this OS, but they aren’t pushed. I was able to build OpenSUSE images locally for arm64 without any issues. What’s the plan for OpenSUSE?

phlogistonjohn:
I'm in favor of starting small. Let's handle the fedora images and then move on to opensuse later. If they're easy enough to handle, I'm in favor of doing it.

Go ahead and squash commits as you see fit.

abachmann:
I extended the GH workflow to build Fedora nightly images as well. For testing, I added a separate commit for the opensuse images to check if they are built in the pipeline. Should I create a new merge request for the opensuse changes?

phlogistonjohn:
@mergifyio rebase

phlogistonjohn (approved):

phlogistonjohn:
mergify rules require a 2nd approval (since this is coming from a contributor and not a maintainer) - @obnoxxx @anoopcs9 @gd - thanks!
I'd like not to have to override mergify :-)

obnoxxx (approved):
LGTM now.
Thanks a lot  for your good contribution and patience with our review requests, @abachmann !

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
Make check fails due to type checking errors caused by the commit:
```
orchestrator/__init__.py:3: note: In module imported here:
orchestrator/module.py: note: In member "_upgrade_status" of class "OrchestratorCli":
orchestrator/module.py:2367: error: Incompatible return value type (got "str", expected "HandleCommandResult")  [return-value]
Found 1 error in 1 file (checked 32 source files)
```

FWIW, it's not particularly difficult to get tox based tests to run on your local machine. I recommend running them prior to making PRs. The flake8 and mypy checks are particularly helpful - they catch a lot little mistakes like this before using CI resources!

PR #181, opened 2024-09-17
images/server: make it easier to do custom package builds

phlogistonjohn:
Creating a custom "devbuilds" container with custom ceph packages and/or custom samba packages was somewhat annoying. Tweak the install-packages.sh to try and make it slightly easier to do one of these builds.


obnoxxx (changes requested):
shellcheck fails in the CI. please fix

obnoxxx (approved):
LGTM

PR #185, opened 2024-11-22
docs: add a script to do the tedious work of fetching digests

phlogistonjohn:
Add a copy-and-paste-able shell script that one can run to generate the downloads section of the release notes. This saves the effort of copy and pasting digests out of the quay.io ui.

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -105,23 +105,51 @@ on GitHub (beyond the sources automatically provided there). Instead we add
 a Downloads section that notes the exact tags and digests that the images can
 be found at on quay.io.
 
-Use the following partial snippet as an example:
-```
-Images built for this release can be obtained from the quay.io image registry.
+The downloads section can be generated using the following shell script:
+```bash
+#!/usr/bin/env bash
+# Requires `skopeo` and `jq` to be installed.
 
-### samba-server
-* By tag: quay.io/samba.org/samba-server:v0.3
-* By digest: quay.io/samba.org/samba-server@sha256:09c867343af39b237230f94a734eacc8313f2330c7d934994522ced46b740715
-### samba-ad-server
-* By tag: quay.io/samba.org/samba-ad-server:v0.3
-* By digest: quay.io/samba.org/samba-ad-server@sha256:a1d901f44be2af5a516b21e45dbd6ebd2f64500dfbce112886cdce09a5c3cbd5
-```
-... and so on for each image that was pushed earlier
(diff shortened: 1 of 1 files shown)


docs/release-process.md
```diff
@@ -105,23 +105,51 @@ on GitHub (beyond the sources automatically provided there). Instead we add
...
-
-### samba-server
-* By tag: quay.io/samba.org/samba-server:v0.3
-* By digest: quay.io/samba.org/samba-server@sha256:09c867343af39b237230f94a734eacc8313f2330c7d934994522ced46b740715
-### samba-ad-server
-* By tag: quay.io/samba.org/samba-ad-server:v0.3
-* By digest: quay.io/samba.org/samba-ad-server@sha256:a1d901f44be2af5a516b21e45dbd6ebd2f64500dfbce112886cdce09a5c3cbd5
+The downloads section can be generated using the following shell script:
+```bash
+#!/bin/bash
```
obnoxxx, line 110:
```suggestion
#!/usr/bin/env bash
```

obnoxxx (changes requested):
request: use env in  script shebang line instead   of direct path to bash.

idea/suggestion: instead of having a copy-and-paste-able script in the doc, how about having an actual script in the repo?

docs/release-process.md
phlogistonjohn: I'm curious, why do you recommend that?

anoopcs9: I can only think of the [portability](https://en.wikipedia.org/wiki/Shebang_(Unix)#Portability) use case. But this brings in a level of indirection. Given that I don't think we have considered the portability aspect elsewhere in samba-container repository and this hasn't been a problem till now.

just my 2 cents.

anoopcs9 (dismissed):
lgtm.

docs/release-process.md
phlogistonjohn: OK, I made the change even though I didn't think it was very necessary. I just figured it would be less work to make the change than to argue it ;-)

anoopcs9 (approved):
still lgtm.

anoopcs9:
@Mergifyio rebase

obnoxxx (approved):
LGTM.

Tried it locally too with the version tag as cmgline arg.

two thoughts that can be addressed in follow-up changes:

* put the script into the repo as an actual script under `hack/`
* mention in the doc how to call it (version tag as first and only argument.

PR #189, opened 2025-01-03
images/server: Enable resilientstorage only for non-nightly builds

anoopcs9:
Resilient Storage is only required when ctdb packages are installed from standard repositories. Therefore avoid enabling it for container builds with samba nightly rpms.


phlogistonjohn (approved):

obnoxxx (approved):
LGTM

PR #191, opened 2025-01-07
build-image enhancements

phlogistonjohn:
New features for hack/build-image:
* A --archive option for writing image bundles to a given directory
* A --load option for reading image bundles from a given directory
* A new --combined option that will create and manage image indexes (manifests)
* Ability to disable unqualified tag names with --no-unqualified

This is fairly large group of changes to prepare the build-image script for multi-arch management.
It tries to enable new workflows such as building images on disparate hosts with varying architectures, sharing those builds with --archive and gathering them with --load, before pushing them with --combined to create multi-arch indexes.

One workflow example goes a bit like:
```
# on builder node 1 (amd64)
# build an amd64 samba server image and then save it to a shared archive dir
$ ./hack/build-image  -i quay.io/phlogistonjohn/samba-server:default-fedora-amd64 --image-condition=rebuild  --archive=image_archive

# on builder node 2 (arm64)
# build an arm64 samba server image and then save it to a shared archive dir
$ ./hack/build-image  -i quay.io/phlogistonjohn/samba-server:default-fedora-arm64 --image-condition=rebuild  --archive=image_archive

# on builder node 3 (arbitrary)
# list contents of shared archive dir
$ ls image_archive
samba-server.default-fedora-amd64.tar  samba-server.default-fedora-arm64.tar
# import images from shared archive dir
./hack/build-image -i quay.io/phlogistonjohn/samba-server:default-fedora-amd64  -i quay.io/phlogistonjohn/samba-server:default-fedora-arm64 --load=image_archive

# automatically create image indexes (manifests) and push the FQINs and indexes to quay.io
./hack/build-image -i quay.io/phlogistonjohn/samba-server:default-fedora-amd64  -i quay.io/phlogistonjohn/samba-server:default-fedora-arm64 --combined --push --push-selected-tags=mixed  --push-state=auto-index
```

### DIFF
--- a/hack/build-image
+++ b/hack/build-image
@@ -351,6 +351,10 @@ def container_push(cli, push_name, manifest=False):
         args = [eng, "manifest", "push", "--all"]
     else:
         args = [eng, "push"]
+    if "podman" in eng:
+        # very strange bug? sometimes it fails to push without the format
+        # option. but I swear it also does work without it some times
+        args.append("--format=oci")
     args.append(push_name)
     run(cli, args, check=True)
 


hack/build-image
```diff
@@ -979,6 +991,14 @@ def main():
             " will be created."
         ),
     )
+    parser.add_argument(
+        "--unqualified",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help=(
+            "Specify if image tags like nightly or latest" " will be created."
```
anoopcs9, line 1003:
``` suggestion
            "Specify if image tags like 'nightly' or 'latest' will be created."
```

hack/build-image
```diff
@@ -733,15 +739,25 @@ class QMatcher:
...
     """Command to push images."""
-    if cli.push_state == "rebuild":
-        build(cli, target)
-    if cli.push_state == "exists":
-        try:
-            container_id(cli, target)
-        except subprocess.CalledProcessError:
-            build(cli, target)
+    is_index = isinstance(target, TargetIndex)  # is it a manifest push?
+    _autobuild(cli, target, is_index=is_index)
```
anoopcs9, line 776:
```suggestion
    _autobuild(cli, target)
```

The above will do, right?

anoopcs9 (commented):
Few comments on commits excluding `hack/build-image: add --combined option for building indexes`.

hack/build-image
```diff
@@ -304,6 +305,16 @@ def container_id(cli, target):
...
+def maybe_container_id(cli, target):
+    """Same as container_id but returns None if no match for the target was
+    found.
+    """
+    try:
+        return container_id(cli, target)
+    except subprocess.CalledProcessError:
+        return None
+
+
```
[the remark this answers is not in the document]
phlogistonjohn: I'm not sure what you mean. did you see `hack/build-image: clean up logic around auto-building on push & archive` ?  If the q is not related to doing something like that patch, can you please elaborate on your question.

hack/build-image
```diff
@@ -590,6 +615,16 @@ def print_tags(cli, target):
...
+def _kwbind(func, key, conversion=None):
+    """Attach an argument value to a command-linked argument."""
+
+    def _capture(arg_value):
+        value = conversion(arg_value) if conversion else arg_value
+        return functools.partial(func, **{key: value})
+
+    return _capture
+
+
```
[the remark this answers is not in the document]
phlogistonjohn: I'm not sure what you mean. I added _kwbind to allow creating calls like `--archive=<dir>` instead of `--archive  --archive=dir=<dir>`. It is kind of silly but I didn't want to change the whole CLI to use subcommands because I dislike changing the makefile (among other reasons).

hack/build-image
phlogistonjohn: Good catch. Will fix.

hack/build-image
phlogistonjohn: `is_index` is passed as `manifest` arg of container_push below.  Making `_autobuild` take this param was not strictly necessary but for some reason I felt compelled to not repeat the isinstance check :shrug:

hack/build-image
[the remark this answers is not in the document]
anoopcs9: Ok. I think I get it now.

hack/build-image
```diff
@@ -207,45 +215,122 @@ def container_build(cli, target):
...
+        add_args.append(
+            "--annotation=org.samba.samba-container.distro=" f"{img.distro}"
+        )
+        add_args += [target.image_name(), img.image_name()]
+        run(cli, add_args, check=True)
+    # apply additional tag names
+    for tname in target.all_names(baseless=cli.without_repo_bases):
+        tag_args = [eng, "tag", target.image_name(), tname]
+        if target.image_name() != tname:
+            run(cli, tag_args, check=True)
```
anoopcs9, line 302: As per your example from description, is this the place which decides on the tag under which manifests are grouped? In the light of your example how do we arrive at the tag name `default-fedora`?

phlogistonjohn: Not really, that is now handled by the BuildRequest class - in particular the _expand_special_tags (and add_special_tags) and  _build_indexes methods.

The tag `default-fedora` for an image is considered a distro qualified tag. For an index it's the fully qualified tag (as the arch belongs w/in the index). The fully qualified tags are determined when each target class is constructed. So if you pass --combined, we walk through the list of requested images and if FQINs exist we'll derive  the index tags from those (in _build_indexes and TargetIndex.from_image).

anoopcs9: > Not really, that is now handled by the BuildRequest class - in particular the _expand_special_tags (and add_special_tags) and _build_indexes methods.

Ok.

> The tag `default-fedora` for an image is considered a distro qualified tag. For an index it's the fully qualified tag (as the arch belongs w/in the index). The fully qualified tags are determined when each target class is constructed. So if you pass --combined, we walk through the list of requested images and if FQINs exist we'll derive the index tags from those (in _build_indexes and TargetIndex.from_image).

But isn't `default-fedora` different from what we had earlier categorized as distro qualified tag? _add_special_tags()_ used to create `fedora-latest` for default package source and `fedora-nightly` for nightly package source as distro qualified tags. With this change _tag_name()_ kind of reverses the tag name parts and doesn't use `latest` for default package source.

hack/build-image
```diff
@@ -724,8 +1044,31 @@ def main():
...
+            " for a given FQIN. Requires FQIN to already exist locally."
+        ),
+    )
+    behaviors.add_argument(
+        "--rebuild",
+        action="store_const",
+        dest="main_action",
+        const=rebuild,
+        help="Always rebuild images even if matching images exist locally.",
+    )
```
anoopcs9, line 1058: How is this different from `--image-condition=rebuild`? In other words can we live with either one of them?

hack/build-image
phlogistonjohn: I'm probably not explaining it very well. I request that you please try it out in practice rather than me going back and forth trying to poorly explain it. If you think something is wrong, please do speak up, but I fear the more I try to explain the code rather than just read it - I will confuse things more.

hack/build-image
phlogistonjohn: Image condition only affects commands like push or archive that expect an image and allow it to rebuild or not. --rebuild is a full blown top-level command.

If I had a time machine I'd do the cli different, but I don't want to completely redo it now. Especially since that will break the existing consumers and my own workflows.

anoopcs9: Ah..I get it now, thanks.

hack/build-image
anoopcs9: I went through the steps to create `samba-server:default-fedora-amd64` and `samba-server:default-fedora-arm64`:
```
. . .
Successfully tagged localhost/samba-server:fedora-latest
Successfully tagged localhost/samba-server:default-fedora-amd64
Successfully tagged quay.io/samba.org/samba-server:fedora-latest
Successfully tagged quay.io/samba.org/samba-server:latest
Successfully tagged quay.io/samba.org/samba-server:default-fedora-amd64
46c6d74fa62c487e96f1c3685e453539b8771d722378b6a34731c4c4351ec4bf
. . .
Successfully tagged localhost/samba-server:fedora-latest
Successfully tagged localhost/samba-server:default-fedora-arm64
Successfully tagged quay.io/samba.org/samba-server:fedora-latest
Successfully tagged quay.io/samba.org/samba-server:latest
Successfully tagged quay.io/samba.org/samba-server:default-fedora-arm64
480500c271df8b226022ab125eff1f4bd8e0d6985c147a38ea6531ec5eb80494
. . .
```

and then executed the following command:
```
$ hack/build-image -i quay.io/anoopcs/samba-server:default-fedora-arm64 -i quay.io/anoopcs/samba-server:default-fedora-amd64 --combined --push --push-selected-tags=mixed --push-state=auto-index
```

The above command resulted in two different tags with 2 manifests each as follows:
```
$ podman image ls | grep anoopcs
quay.io/anoopcs/samba-server                  default-fedora          20191b9fb0da  6 minutes ago   1.05 kB
quay.io/anoopcs/samba-server                  latest                  20191b9fb0da  6 minutes ago   1.05 kB
quay.io/anoopcs/samba-server                  default-fedora-amd64    46c6d74fa62c  56 minutes ago  472 MB
quay.io/anoopcs/samba-server                  default-fedora-arm64    480500c271df  57 minutes ago  581 MB
```

Earlier we created the distro qualified tag as `fedora-latest` and now it gets tagged as `default-fedora`. I hope this new naming happens within _tag_names()_.

PR #195, opened 2025-01-09
common:  implement an explicit preference order in sambacc installation

obnoxxx:
this explicitly implements a preference order for sambacc
    custom installation sources like so:

    local repo ->  -> rpm ->  wheel   -> copr.

 It makes the output a bit more verbose while at it.
overall, the behavior of the sambacc installation is made a bit more consistent and predictable

### DIFF
--- a/images/common/install-sambacc-common.sh
+++ b/images/common/install-sambacc-common.sh
@@ -30,28 +30,42 @@ install_sambacc() {
     done
 
 
-    local action=install-from-copr-repo
-    if [ "${#wheels[@]}" -gt 1 ]; then
-        echo "more than one wheel file found"
-        exit 1
-    elif [ "${#wheels[@]}" -eq 1 ]; then
-        action=install-wheel
+    local action="" 
+    echo "INFO: determining sambacc installation source." 
+    if [ "${#wheels[@]}" -gt 0 ]; then
+        echo "INFO: wheel found"
+        if [ "${#wheels[@]}" -gt 1 ]; then
+            echo "ERROR: more than one wheel file found"
+            exit 1
+        elif [ "${#wheels[@]}" -eq 1  ]; then
+            action=install-wheel
+        fi
+    fi
+    if [ "${#rpmfiles[@]}" -gt 0 ]; then
+        echo "INFO: rpm file found"
+        if [ "${#rpmfiles[@]}" -gt 1 ]; then
+            echo "ERROR: more than one sambacc rpm file found"
+            exit 1
+        elif [ "${#rpmfiles[@]}" -eq 1 ]; then
+            action=install-rpm
+        fi
+    fi
+    if [ "${#repofiles[@]}" -gt 0 ]; then
+       echo  "INFO: repo file found"
+       if [ "${#repofiles[@]}" -gt 1 ]; then
+           echo "ERROR: more than one repo file found"
+           exit 1
+       elif [ "${#repofiles[@]}" -eq 1 ]; then
+           action=install-from-repo
+       fi
+    fi
+    if [ -z "${action}" ]; then
+
+        echo "INFO: no local sambacc installation source found. falling back to copr install."
+        action=install-from-copr-repo
     fi
 
-    if [ "${#rpmfiles[@]}" -gt 1 ]; then
-        echo "more than one sambacc rpm file found"
-        exit 1
-    elif [ "${#rpmfiles[@]}" -eq 1 ]; then
-        action=install-rpm
-    fi
-
-    if [ "${#repofiles[@]}" -gt 1 ]; then
-        echo "more than one repo file found"
-        exit 1
-    elif [ "${#repofiles[@]}" -eq 1 ]; then
-        action=install-from-repo
-    fi
-
+    echo "INFO: selected installation method: '${action}'"
     if [ -z "${DEFAULT_JSON_FILE}" ]; then
         echo "DEFAULT_JSON_FILE value unset"
         exit 1


obnoxxx:
@phlogistonjohn : the requests you had on a previous version of this  patch in #193  have all been addressed

phlogistonjohn (changes requested):
I think this has (unintentionally) inverted the priority of the various installation sources. Previously, if it found a wheel and then an rpm it would "prefer" the RPM. Now, AFAICT it will find a wheel and stop looking for anything else, implicitly preferring the wheel.

I strongly recommend undoing the restructuring of the if blocks so that every if block still gets visited and the various log lines can be emitted. Then after the sequence of if-blocks you can log the state of `action` to indicate what the 'preferred' action is going to be.

obnoxxx:
@phlogistonjohn  wrote:
> I think this has (unintentionally) inverted the priority of the various installation sources. Previously, if it found a wheel and then an rpm it would "prefer" the RPM. Now, AFAICT it will find a wheel and stop looking for anything else, implicitly preferring the wheel.

Indeed, this analysis is correct and it was not intentional.

> 
> I strongly recommend undoing the restructuring of the if blocks so that every if block still gets visited and the various log lines can be emitted. Then after the sequence of if-blocks you can log the state of `action` to indicate what the 'preferred' action is going to be.

I have changed it this way now, making the original preference order explicit instead of reverting it.

images/common/install-sambacc-common.sh
```diff
@@ -31,26 +31,48 @@ install_sambacc() {
...
-        exit 1
-    elif [ "${#rpmfiles[@]}" -eq 1 ]; then
-        action=install-rpm
+    if [ "${#rpmfiles[@]}" -gt 0 ]; then
+        echo "INFO: rpm file found"
+        if [ "${#rpmfiles[@]}" -gt 1 ]; then
+            echo "ERROR: more than one sambacc rpm file found"
+            exit 1
+        elif [ "${#rpmfiles[@]}" -eq 1 ]; then
+            echo "sambacc rpm found. preferring  rpm install."
```
phlogistonjohn, line 52: Extra space between preferring and rpm: `preferring  rpm`.

images/common/install-sambacc-common.sh
```diff
@@ -31,26 +31,48 @@ install_sambacc() {
 
 
     local action=install-from-copr-repo
-    if [ "${#wheels[@]}" -gt 1 ]; then
-        echo "more than one wheel file found"
-        exit 1
-    elif [ "${#wheels[@]}" -eq 1 ]; then
-        action=install-wheel
+    local source_found=0
```
phlogistonjohn, line 34: Why not just leave `action` unset and then do a test like `[ -z $action ]` for the other if block? No need for a 2nd variable.

images/common/install-sambacc-common.sh
```diff
@@ -31,26 +31,48 @@ install_sambacc() {
...
+    if [ "${#wheels[@]}" -gt 0 ]; then
+        echo "INFO: wheel found"
+        if [ "${#wheels[@]}" -gt 1 ]; then
+            echo "ERROR: more than one wheel file found"
+            exit 1
+        elif [ "${#wheels[@]}" -eq 1  ]; then
+            echo  "sambacc wheel found. Installing wheel."
```
phlogistonjohn, line 41: I think these lines (like `echo  "sambacc wheel found. Installing wheel."` and ` echo "sambacc rpm found. preferring  rpm install."` really ought to be saved until *after* the if statements have all been processed. Otherwise, we will get results where it might print something like:

```
INFO: wheel found
sambacc wheel found. Installing wheel.
INFO: rpm file found
sambacc rpm found. preferring  rpm install.
INFO: repo file found"
sambacc repo file found. preferring  install from local yum repo.
```

If I didn't know the code well, I'd find that confusing. I think the `INFO:`  lines are good, they're true. But I'd rather not say anything about (preferred) install method until after the final assignment of `action`.

images/common/install-sambacc-common.sh
```diff
@@ -31,26 +31,48 @@ install_sambacc() {
...
+       elif [ "${#repofiles[@]}" -eq 1 ]; then
+           echo "sambacc repo file found. preferring  install from local yum repo."
+           action=install-from-repo
+           source_found=1
+       fi
     fi
+        if [ "${source_found}" = "0" ]; then
+            echo "INFO: no local sambacc installation source found. falling back to copr install."
+            action=install-from-copr-repo
+        fi
```
phlogistonjohn, line 71: This if-block is overly indented.

obnoxxx: fixed

images/common/install-sambacc-common.sh
obnoxxx: agreed. Will address tomorrow.

obnoxxx: done

images/common/install-sambacc-common.sh
obnoxxx: good point! done.

images/common/install-sambacc-common.sh
obnoxxx: gone, as is the entire message.

images/common/install-sambacc-common.sh
```diff
@@ -30,27 +30,44 @@ install_sambacc() {
...
+           exit 1
+       elif [ "${#repofiles[@]}" -eq 1 ]; then
+           action=install-from-repo
+       fi
     fi
+       if [ -z "${action}" ]; then
+
+           echo "INFO: no local sambacc installation source found. falling back to copr install."
+           action=install-from-copr-repo
+       fi
```
phlogistonjohn, line 66: Looking a lot better now, but this section is still overly indented.

obnoxxx: really fixed now

images/common/install-sambacc-common.sh
```diff
@@ -30,28 +30,45 @@ install_sambacc() {
...
-    elif [ "${#repofiles[@]}" -eq 1 ]; then
-        action=install-from-repo
+        echo "INFO: no local sambacc installation source found. falling back to copr install."
+        action=install-from-copr-repo
     fi
 
+    echo "INFO: selected installation method: '${action}'"
+
+
+
```
phlogistonjohn, line 71: Now there's a bunch new whitespace here.

phlogistonjohn (changes requested):

images/common/install-sambacc-common.sh
obnoxxx: gone

phlogistonjohn (approved):
lgtm, thanks

anoopcs9 (approved):

PR #198, opened 2025-03-17
hack: conditionally set samba specifics value based on source

phlogistonjohn:
Conditionally set the samba specifics value for ctdb_rados_mutex_skip_reg based on source. When the source is a samba master build (nightly, devbuilds) we want to set this specifics flag.

Reformat source of `hack/build-image` with black as I noticed it had many small issues when reading it over today.


avanthakkar (approved):

spuiuk (approved):
ACK

PR #203, opened 2025-05-20
docs: add dnf repo support to scc builder example

phlogistonjohn:
In the development doc the *Build Stage* section documents a method for building sambacc packages that will become part of the final samba-server image. This is very handy for practical testing of sambacc changes. But the document is a bit out of date in that the pure "find rpms and install them" method doesn't work well when the extras are needed.
Update the file to document creating a yum/dnf repo and .repo file such that dnf will be used with the sambacc target package in order for `Recommends` to be considered during the install.


anoopcs9 (approved):
lgtm, thanks.

PR #210, opened 2025-07-15
server: support more customization

phlogistonjohn:
To perform practical tests with a bunch of incomplete work I needed to combine custom dev builds of ceph, and samba, and sambacc.  This series implements some work to make creating very custom image builds easier. The parts I could cleanly implement quickly at least.

These changes should have no impact on typical non-custom builds.


synarete (approved):
lgtm

PR #212, opened 2025-07-25
images/server: Improvements to install-pacakges.sh

anoopcs9:
Actually we could build on `--install-custom-repo` rather than new `--custom-repos`. Also in the absence of _forcedevbuilds_ value for `--package-selection`, by default we failed to install some important packages for _custom_ package source.


anoopcs9 (commented):
> I like the general direction this is going but I have some quibbles with the way the names a little confused WRT to plurals. But one issues is do we want to change all the variable names or is that too much of a pain.

I've made changes to indicate the plural nature both internally and for options.

> I forgot to mention: with regards to the naming, since we've added long options I think it'd be OK to have alises if we want like --install-custom-repo and --custom-repos could store to the same variable IF it makes things clearer and easier to use.

I hope the rename to `--install-custom-repos` is good enough for now.

> Also also, since #210 got merged the script relies on build args and the build args map to the existing options, so I think both ought to be updated to match (probably in one PR, but not necessarily one patch)

Can you please check the updated version?

images/server/install-packages.sh
```diff
@@ -128,7 +128,7 @@ case "${install_packages_from}" in
     ;;
     custom-repos)
         get_custom_repos "${install_custom_repos}"
-        package_selection=${package_selection:-custom}
+        package_selection=${package_selection:-custom-repos-devbuilds}
```
phlogistonjohn, line 131: No complaint with this change, but I had figured people (usually me) doing custom builds would also be setting the appropriate `--package-selection` value each time. So that's why I hadn't changed it previously.

phlogistonjohn (approved):
lgtm thanks.

phlogistonjohn:
@anoopcs9 are you OK to merge this? Either one of us can follow up on the ceph repo option aspect in a follow up pr. Do you want a 2nd review?

images/server/install-packages.sh
```diff
@@ -12,8 +12,13 @@ need_curl() {
 
 get_custom_repos() {
     if [[ -n $1 ]]; then
+        urls=$1
         need_curl
-        for url in $1; do
+        if [[ $(wc -w <<< "$1") -eq 1 || $install_custom_ceph -ne 1 ]]; then
+            get_distro_ceph_repo
+            urls=${1%% *}
```
phlogistonjohn, line 19: What's this line do?

images/server/install-packages.sh
```diff
@@ -95,6 +97,7 @@ if [[ "$1" =~ ^--.+$ ]]; then
             --samba-version-suffix=*) samba_version_suffix="${arg/*=/}" ;;
             --install-custom-repos=*) install_custom_repos="${arg/*=/}" ;;
             --package-selection=*) package_selection="${arg/*=/}" ;;
+            --install-custom-ceph) install_custom_ceph=1 ;;
```
phlogistonjohn, line 100: I am struggling to follow this the naming scheme / flow.

If we pass exactly one custom repo _or_ `install_custom_ceph` is not `1` (say 0) it will enable distro-ceph/chacra. Passing `--install-custom-ceph` sets `install_custom_ceph=1`.  

So it seems to me that `--install-custom-ceph` means "I (the user) am providing a ceph repo", right? It seems a bit confusing to me because I might take it to mean `the script is (or should be) providing a ceph repo`. 

How does `--ceph-from-custom` strike you? I tried to make it clear that ceph is coming from the list of custom repos *you, the user* are supplying. I'm open to other suggestions, but I think that `--install-custom-ceph` will just force me to read the code every time I need it to remind me how it works.

phlogistonjohn (changes requested):

anoopcs9:
Updated once again with two additional commits (not related to custom repo management).

phlogistonjohn (dismissed):
lgtm

images/server/install-packages.sh
xhernandez: This won't work if, for some reason, the string starts with a space. Why not use bash arrays ?

```bash
urls=(${1})
...
if [[ ${#urls[@]} -eq 1 || $ceph_from_custome -ne 1 ]]; then
    ...
    urls=(${urls[0]})
fi
for url in "${urls[@]}"; do
...
```

images/server/install-packages.sh
```diff
@@ -10,20 +10,28 @@ need_curl() {
...
+            urls=${1%% *}
         fi
-        dest="/etc/yum.repos.d/${u}-${fname}"
-    done
-    need_curl
-    curl -L "$url" -o "$dest"
+        for url in $urls; do
+            fname="$(basename "$url")"
+            dest="/etc/yum.repos.d/${fname}"
+            while [ -e "$dest" ]; do
```
xhernandez, line 24: If the first iteration of the loop fails (i.e. the file with the hash prefixed still exists), this will become an infinite loop because the generated name will be the same.

I think this could be converted to a simple `if`. If the name with the hash exists, then it has been downloaded from the same place, so the file we'll download is most likely the same.

images/server/install-packages.sh
```diff
@@ -10,20 +10,28 @@ need_curl() {
...
-    while [ -e "$dest" ]; do
-        u="$(echo "${url}" | sha256sum | head -c12)"
-        if [ -z "$u" ]; then
-            echo "failed to uniquify repo file name"
-            exit 2
+get_custom_repos() {
+    if [[ -n $1 ]]; then
+        urls=$1
+        need_curl
+        if [[ $(wc -w <<< "$1") -eq 1 || $ceph_from_custom -ne 1 ]]; then
```
xhernandez, line 17: Why we just install the first repo if `ceph_from_custom` is not specified ? if this option is not specified, only one additional repo can be installed ?

Also, if `ceph_from_custom` is specified and we have a single custom repo, why we install the standard ceph repo ? shouldn't it be installed from one of the custom repos ?

images/server/install-packages.sh
```diff
@@ -10,20 +10,28 @@ need_curl() {
...
+            get_distro_ceph_repo
+            urls=${1%% *}
```
xhernandez, line 19: This won't work if, for some reason, the string starts with a space. Why not use bash arrays ?

```bash
urls=(${1})
...
if [[ ${#urls[@]} -eq 1 || $ceph_from_custome -ne 1 ]]; then
    ...
    urls=(${urls[0]})
fi
for url in "${urls[@]}"; do
...
```

images/server/install-packages.sh
```diff
@@ -12,8 +12,13 @@ need_curl() {
 
 get_custom_repos() {
     if [[ -n $1 ]]; then
+        read -ra urls <<< "$1"
         need_curl
-        for url in $1; do
+        if [[ ${#urls[@]} -eq 1 || $ceph_from_custom -ne 1 ]]; then
+            get_distro_ceph_repo
+            urls=("${urls[0]}")
```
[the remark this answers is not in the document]
xhernandez: If more than one repo could be related to Samba or any other non-ceph component, then we cannot do that.

Why not just have two options, one to specify an optional ceph repo instead of the default one, something like `--ceph-repo`, and another to add as many additional repos as required, like `--extra-repos`

phlogistonjohn (approved):
thanks. looks good to me

phlogistonjohn:
mergify!! :shakingfist:

phlogistonjohn:
Mergiyf is complaining:
> The branch protection setting Require branches to be up to date before merging is not compatible with max_parallel_checks>1, queue_conditions != merge_conditions and must be unset.

Something to follow up with later as I am very disappointed in mergify lately. Manually merging now.

PR #215, opened 2025-08-12
images/server: Fix ceph shaman base url parameters

anoopcs9:
The parameters for the base shaman url that was used to determine the chacra repository link did not have `status=ready` resulting in invalid chacra urls as part of the json response. Make sure that we fetch with `status=ready` so as to create repo files with valid base url.


spuiuk (approved):
ACK

PR #221, opened 2025-09-08
images: update base fedora image version to 42

phlogistonjohn:
Now we finally released v0.7 we can bump this up again.


synarete (approved):
LGTM

PR #223, opened 2025-09-16
images/server: Updates to package installation

anoopcs9:
- Use Samba v4.23 for `ceph20`.
- Switch to Storage SIG ceph tentacle packages for `ceph20` on CentOS.
- Generalize a switch case for CentOS based package selection.

### DIFF
--- a/images/server/install-packages.sh
+++ b/images/server/install-packages.sh
@@ -198,12 +198,8 @@ case "${package_selection}-${OS_BASE}" in
     *-fedora|allvfs-*)
         samba_packages+=(samba-vfs-cephfs samba-vfs-glusterfs ctdb-ceph-mutex)
     ;;
-    *devbuilds-centos|forcedevbuilds-*|stable-*)
-	# Enable libcephfs proxy for dev builds
+    *-centos|forcedevbuilds-*)
         support_packages+=(libcephfs-proxy2)
-	# Fall through to next case
-    ;&
-    nightly-centos|default-centos)
         samba_packages+=(samba-vfs-cephfs samba-vfs-glusterfs ctdb-ceph-mutex)
         # these packages should be installed as deps. of sambacc extras
         # however, the sambacc builds do not enable the extras on centos atm.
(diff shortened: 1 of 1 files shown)


xhernandez (approved):

anoopcs9:
Forgot to remove the shellcheck waiver ([diff](https://github.com/samba-in-kubernetes/samba-container/compare/56b349a5721052fa3d2e8d8f621ea762fde666d2..b02c885e1ff463b918f7831a19e2487e2d9df065)).

images/server/install-packages.sh
```diff
@@ -74,7 +74,6 @@ get_sig_samba_repo() {
     fi
 }
 
-# shellcheck disable=SC2120
```
xhernandez, line 77: Does this mean that it's expected that all calls to this function will explicitly provide an argument ? if so, I think we should also remove the `if [[ -z $1 ]]; then`.

xhernandez: Ah, ok. Thanks.

phlogistonjohn:
> * Switch to Storage SIG ceph tentacle packages for `ceph20` on CentOS.

I'm not in favor of this. We ship as part of ceph when the smb mgr module is used. The ceph containers are built using the packages that appear in shaman/chacra we should use the same set of packages to be as close to ceph upstream as possible.

phlogistonjohn:
I don't know how much they differ and that's part of the issue. I do know how the RPMs produced by the ceph CI work and where all that is managed. My main point is that the images we build specifically to be consumed by ceph users should hew as closely to what the ceph images have - and that implies the packages produced by ceph directly not the storage sig IMO.

anoopcs9:
> I don't know how much they differ and that's part of the issue. I do know how the RPMs produced by the ceph CI work and where all that is managed. My main point is that the images we build specifically to be consumed by ceph users should hew as closely to what the ceph images have - and that implies the packages produced by ceph directly not the storage sig IMO.

I have confidence in how Ceph releases are packaged in Storage SIG. It is actively maintained. There's also the problem with the current tentacle RPM builds that we consume via shaman infrastructure which is based on the moving `tentacle` branch and NOT the actual `tentacle-release` branch. It is in our best interest to ship released versions and not something which might become the next.

PR #247, opened 2026-04-20
images/common: Remove a COPR chroot workaround for CentOS Stream 9

anoopcs9:
Remove the hack to enable a different chroot for the dependencies of sambacc as we have those built and available in the default chroot in sambacc and sambacc-extras-deps COPR repositories.


phlogistonjohn (approved):

spuiuk (approved):
ACK

PR #250, opened 2026-05-06
workflows: disable fail-fast on build matrixes

phlogistonjohn:
Fail-fast is nice if your failures are largely driven by invalid code but when the failures are caused by infra out of your control it can mask other issues. Disable fail-fast on our build jobs so we can monitor them independently while dealing with infra problems in ceph.

### DIFF
--- a/.github/workflows/container-image.yml
+++ b/.github/workflows/container-image.yml
@@ -42,6 +42,7 @@ jobs:
     # Reminder: the nightly-server images consume nightly samba rpm builds
     # it is not *just* an image that gets built nightly
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


anoopcs9 (commented):
What about the `test-server` and `test-ad-server-kubernetes` jobs? Disabling fail-fast can also help those jobs, right?

phlogistonjohn:
> What about the test-server and test-ad-server-kubernetes jobs? Disabling fail-fast can also help those jobs, right?

Yeah I suppose the same principle holds and its good to see statuses independently of the sibling test jobs too. Done.

anoopcs9 (approved):
lgtm, thanks.

PR #254, opened 2026-06-15
[RFC] images: exclude non-functioning EPEL build of pykmip

phlogistonjohn:
Fixes: #253 

EPEL 10 includes a build of python3-pykmip that is considered newer than the build from our COPR but didn't fix the bug(s) that our build does. Exclude that unwanted build from the container build via DNF config.


phlogistonjohn:
No one seems in favor of this approach. Closing.

